PR #1, opened 2018-07-23
tracer option to exclude objc methods

fopina:
It's always useful to have the `exclude` option on each ApiResolver because of wildcards :)


oleavr:
Just released in 1.1.0. Cheers!

PR #4, opened 2018-08-25
Add frida-trace -q (--quiet) to omit formatting

trufae:


frida_tools/tracer.py
```diff
@@ -14,6 +14,7 @@
 
 from frida_tools.model import Module, Function, ModuleFunction, ObjCMethod
 
+isQuiet = False
```
oleavr: This should go in `_initialize` and be assigned from the `options` argument, and PEP-8 style i.e. `is_quiet`.

frida_tools/tracer.py
```diff
@@ -32,6 +33,8 @@ def _add_options(self, parser):
             pb = TracerProfileBuilder()
             def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
                 method(value)
+            parser.add_option("-q", "--quiet", help="do not format agents output",
+                    action='callback', callback=process_builder_arg, callback_args=(pb.quiet,))
```
oleavr: Better to use `action='store_true'` and drop the `callback` and `callback_args`.

oleavr:
Committed with minor tweaks as 97243c36f88a003b8678ada96a38576da52c05e1. Thanks! 🤘

PR #5, opened 2018-08-28
Implement frida-trace -o

trufae:


frida_tools/tracer.py
```diff
@@ -106,7 +110,10 @@ def on_trace_error(self, error):
         def on_trace_events(self, events):
             no_attributes = Style.RESET_ALL
             for timestamp, thread_id, depth, target_address, message in events:
-                if self._quiet:
+                if self._output is not None:
+                    with open(self._output, "a") as fh:
```
dweinstein: won't this have the effect of constantly opening and closing the output for every event? that's a lot of syscalls isn't it?

dweinstein: seems like it would be more efficient to open the file once outside the loop at least, or open it once at the class level and check that the handle is available before writing the message?

dweinstein:
I worked up another approach here #6

trufae:
i have cherry-picked your open->append approch which is more optimal and seems to work reliably even when the app crashes or the frida-trace is killed, which was my concern for not doing it this way in the first turn.

I have fixed the behaviour of `-o` in your PR because when using this flag I expect frida-trace to write all the output into the file and not in the console.

Thanks

frida_tools/tracer.py
```diff
@@ -53,6 +52,7 @@ def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
             parser.add_option("-s", "--include-debug-symbol", help="include DEBUG_SYMBOL", metavar="DEBUG_SYMBOL",
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format agent's output", action='store_true')
+            parser.add_option("-o", "--output", help="dump agent messages to file", metavar="OUTPUT", type='string')
```
oleavr: Perhaps just: `dump messages to file`, as it's probably unnecessary to mention `agent` here. (If so we should drop it for `--quiet` also.)

trufae:
addressed all comments

frida_tools/tracer.py
```diff
@@ -105,16 +110,19 @@ def on_trace_error(self, error):
...
-                    attributes = self._get_attributes(thread_id)
-                    if thread_id != self._last_event_tid:
-                        self._print("%s           /* TID 0x%x */%s" % (attributes, thread_id, Style.RESET_ALL))
-                        self._last_event_tid = thread_id
-                    self._print("%6d ms  %s%s%s%s" % (timestamp, attributes, indent, message, no_attributes))
+            with OutputFile(self._output) as output:
+                for timestamp, thread_id, depth, target_address, message in events:
+                    if output is not None:
+                        output.append(message + "\n")
+                    elif self._quiet:
```
dweinstein: I guess this has to be is not None comparison to be consistent?

trufae:
i can squash all my fixes for proper rebase merging if you want, just tried to keep it like this for proper review

frida_tools/tracer.py
trufae: Seems like store_true makes the python option parsing use None or True instead of False|True. I have fixed this before caching that option inside the object.

oleavr: How about giving it a default value of `False` instead – won't that make it `False|True`?

oleavr:
Timed out. Feel free to reopen at a later time.

frida_tools/tracer.py
trufae: my last commit fixed this already

frida_tools/tracer.py
```diff
@@ -52,7 +52,8 @@ def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.exclude_objc_method,))
             parser.add_option("-s", "--include-debug-symbol", help="include DEBUG_SYMBOL", metavar="DEBUG_SYMBOL",
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
-            parser.add_option("-q", "--quiet", help="do not format agent's output", action='store_true')
+            parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true')
```
```suggestion
            parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
```

frida_tools/tracer.py
```diff
@@ -62,7 +63,11 @@ def _initialize(self, parser, options, args):
             self._tracer = None
             self._targets = None
             self._profile = self._profile_builder.build()
-            self._quiet = options.quiet
+            self._quiet = options.quiet == True
```
```suggestion
            self._quiet = options.quiet
```

frida_tools/tracer.py
```diff
@@ -62,7 +63,11 @@ def _initialize(self, parser, options, args):
...
+            self._output = options.output
+            if self._output is not None:
+                with OutputFile(self._output, truncate=True) as output:
```
oleavr: We should probably just store the `OutputFile` instance on `self` so we don't open and close the file all the time.

frida_tools/tracer.py
```diff
@@ -918,6 +926,29 @@ def _sync_handlers(self, change_id):
...
 
 
+class OutputFile(object):
+    def __init__(self, filename, truncate=False):
+        self.filename = filename
+        self._truncate = truncate
+        self._fd = None
+
+    def __enter__(self):
+        if self.filename:
```
oleavr: This check is not needed.

frida_tools/tracer.py
```diff
@@ -918,6 +926,29 @@ def _sync_handlers(self, change_id):
...
+            self._fd = codecs.open(self.filename, 'ab', 'utf-8')
+            if self._truncate:
+                self._fd.truncate(0)
```
oleavr: As we're not using the truncate option we should remove it.

frida_tools/tracer.py
```diff
@@ -918,6 +926,29 @@ def _sync_handlers(self, change_id):
...
+            return self
+        return None
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if self._fd is not None:
+            self._fd.close()
```
oleavr: This code should probably be moved to a `close()` method that we call from here. That way this object can be used both as a scope-local context manager and as a long-lived instance.

frida_tools/tracer.py
trufae: its being used already

frida_tools/tracer.py
trufae: its needed if you call the constructor with filename=None.. which shouldnt happen, but with dynamic languages everything can break at anytime.

frida_tools/tracer.py
trufae: i introduced this true because quiet can be None, False and True and i want it to be a boolean not a quantum state

trufae: but this is solved by using the default= in the options setup. solved

frida_tools/tracer.py
trufae: yeah, this outputfile class is pretty messy. i have simplified the whole thing. can you review again?

frida_tools/tracer.py
trufae: this is no longer there, early failure in case the file cant be opened is working and the caes to open a file with no name "" is no longer possible now

frida_tools/tracer.py
trufae: but i have refactored the code to not need it

frida_tools/tracer.py
trufae: yes, the enter/exit scope magic is pretty messy. i have just removed that

frida_tools/tracer.py
```diff
@@ -52,7 +52,8 @@ def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.exclude_objc_method,))
             parser.add_option("-s", "--include-debug-symbol", help="include DEBUG_SYMBOL", metavar="DEBUG_SYMBOL",
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
-            parser.add_option("-q", "--quiet", help="do not format agent's output", action='store_true')
+            parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
+            parser.add_option("-o", "--output", help="dump messages to file", metavar="OUTPUT", type='string', default="")
```
```suggestion
            parser.add_option("-o", "--output", help="dump messages to file", metavar="OUTPUT", type='string')
```

frida_tools/tracer.py
```diff
@@ -63,6 +64,8 @@ def _initialize(self, parser, options, args):
             self._targets = None
             self._profile = self._profile_builder.build()
             self._quiet = options.quiet
+            if options.output:
```
```suggestion
            if options.output is not None:
```

frida_tools/tracer.py
```diff
@@ -918,6 +923,18 @@ def _sync_handlers(self, change_id):
...
 
+class OutputFile(object):
+    def __init__(self, filename):
+        self._fd = codecs.open(self.filename, 'wb', 'utf-8')
+
+    def append(self, message):
+        self._fd.write(message)
+        self._fd.flush()
+
+    def __del__(self):
```
oleavr: Should avoid relying on GC behavior and instead close explicitly:

```suggestion
    def close(self):
```

(Let's also move this method above `append()` as methods should be ordered: constructor, "destructor", <anything else>.)

frida_tools/tracer.py
```diff
@@ -106,7 +109,9 @@ def on_trace_error(self, error):
         def on_trace_events(self, events):
             no_attributes = Style.RESET_ALL
             for timestamp, thread_id, depth, target_address, message in events:
-                if self._quiet:
+                if self._output:
```
```suggestion
                if self._output is not None:
```

frida_tools/tracer.py
```diff
@@ -63,6 +64,8 @@ def _initialize(self, parser, options, args):
             self._targets = None
             self._profile = self._profile_builder.build()
             self._quiet = options.quiet
+            if options.output:
+                self._output = OutputFile(options.output)
```
```suggestion
                self._output = OutputFile(options.output)
            else:
                self._output = None
```

frida_tools/tracer.py
trufae: so it's just the string type that have a meaningful defaults? sholdnt we set also 'boolean' type in the -q and drop the default=False?

frida_tools/tracer.py
oleavr: Nice. Even if anything is possible with dynamic languages, I'm a fan of deciding on API contracts that only support exactly what I need and nothing more, and write the code assuming that the contract is followed exactly. (The public API boundary is a different beast though, as users will be making mistake and usability is important, but for internal APIs I try hard to limit options so the code is as readable as possible.)

frida_tools/tracer.py
oleavr: Sweet! Will do.

frida_tools/tracer.py
oleavr: The `optparse` module ensures that `True` will be stored if the user set the option. But it doesn't have a boolean type, at least not in the 2.7 version (didn't check 3.x), so the default is `None` if you don't specify one. So my thinking was that it is slightly simpler to do that there instead of normalizing it afterwards.

frida_tools/tracer.py
oleavr: 👍

frida_tools/tracer.py
oleavr: 👍

frida_tools/tracer.py
```diff
@@ -78,6 +83,8 @@ def _start(self):
         def _stop(self):
             self._tracer.stop()
             self._tracer = None
+            if self._output is not None:
+                self._output.close()
```
oleavr: Should also set it to `None`.

frida_tools/tracer.py
```diff
@@ -63,6 +64,10 @@ def _initialize(self, parser, options, args):
             self._targets = None
             self._profile = self._profile_builder.build()
             self._quiet = options.quiet
```
oleavr: Should only assign `self._output_path` here, setting `self._output` to `None`, and assign it in `_start()`.

trufae:
updated

Issue #7, opened 2018-10-09
Feature: option to disable terminal colors (colorama)


arizvisa:
would be friendly to have an option or environment variable to disable the usage of colorama so that terminal escape sequences aren't emitted. this way colors aren't imposed on the user and they have a choice.

oleavr:
PR welcome!

arizvisa:
it was easier to just remove it entirely. i can submit a PR that gets rid of the colorama dependency, but i figure since you guys did the work to add coloring that it would be contrary to your guys original intentions.

writing code sucks, but if i submit a PR will you actually merge it?

oleavr:
Yes if it's a PR adding an option to disable it.

arizvisa:
PR #8.

arizvisa:
Re-opened so it's more apparent how this was mishandled.

oleavr:
Let's keep tracking this in #8.

PR #8, opened 2018-10-11
Added support for the 'none' terminal type specified via the TERM environment variable.

arizvisa:
This closes issue #7 by allowing the user to specify 'none' as `TERM` in their environment to avoid emitting terminal escape codes.

Explicit usage of escape codes that were being printed were replaced to use the ones defined in colorama. To filter out the escape codes it was noticed that only standard string concatenation was used via the "+" operator, and so this logic was moved into the `self._*print` functions. This changes the semantics of `self._print` so that each individual arg is not space-separated, but since it wasn't used anywhere it shouldn't matter.

---
```
Added support for the 'none' terminal type specified via the TERM environment variable.

frida-tools/application.py:
    Added the 'none' terminal type to set the '_plain_terminal' property.
    Replaced all explicit cursor position escape codes with the respective
        functions in `colorama.Cursor`.
    Added an implementation of ._uprint() to replace the regular calls to
        the `print()`  function.
    Split up all of the string concatenation that is used to emit terminal
        escape codes into separate variable args for `self._*print()`.
    Added filters to `self._*print` to remove terminal escape codes that
        get emitted.

frida-tools/{repl,tracer}.py:
    Split up all of the string concatenation that is used to emit terminal
        escape codes into separate variable args for `self._*print()`.
```

### DIFF
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -328,7 +328,7 @@ class ConsoleApplication(object):
             self._uprint(Style.BRIGHT, message, Style.RESET_ALL)
 
     def _uprint(self, *args):
-        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
+        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith("\033"))
         print(args)
 
     def _print(self, *args, **kwargs):
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -16,7 +16,7 @@ if platform.system() == 'Windows':
     import msvcrt
 
 import colorama
-from colorama import Fore, Style, Cursor
+from colorama import Cursor, Fore, Style
 import frida
 
 
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -314,7 +314,7 @@ class ConsoleApplication(object):
 
     def _clear_status(self):
         if not self._plain_terminal and self._console_state == ConsoleState.STATUS:
-            self._uprint(Cursor.UP(), (80 * " "))
+            self._uprint(Cursor.UP(), 80 * " ")
 
     def _update_status(self, message):
         if self._have_terminal and not self._plain_terminal:
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -333,7 +333,7 @@ class ConsoleApplication(object):
 
     def _print(self, *args, **kwargs):
         # Filter out any args that are escape codes and turn them into a single element tuple
-        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
+        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith("\033"))
         args = [args]
 
         encoded_args = []
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -16,7 +16,7 @@ if platform.system() == 'Windows':
     import msvcrt
 
 import colorama
-from colorama import Fore, Style
+from colorama import Fore, Style, Cursor
 import frida
 
 
@@ -132,6 +132,7 @@ class ConsoleApplication(object):
         self._exit_status = None
         self._console_state = ConsoleState.EMPTY
         self._have_terminal = sys.stdin.isatty() and sys.stdout.isatty() and not os.environ.get("TERM", '') == "dumb"
+        self._plain_terminal = os.environ.get("TERM", '').lower() == "none"
         self._quiet = False
         if sum(map(lambda v: int(v is not None), (self._device_id, self._device_type, self._host))) > 1:
             parser.error("Only one of -D, -U, -R, and -H may be specified")
@@ -308,25 +309,34 @@ class ConsoleApplication(object):
 
     def _on_session_detached(self, reason):
         message = reason[0].upper() + reason[1:].replace("-", " ")
-        self._print(Fore.RED + Style.BRIGHT + message + Style.RESET_ALL)
+        self._print(Fore.RED, Style.BRIGHT, message, Style.RESET_ALL)
         self._exit(1)
 
     def _clear_status(self):
-        if self._console_state == ConsoleState.STATUS:
-            print("\033[A" + (80 * " "))
+        if not self._plain_terminal and self._console_state == ConsoleState.STATUS:
+            self._uprint(Cursor.UP(), (80 * " "))
 
     def _update_status(self, message):
-        if self._have_terminal:
+        if self._have_terminal and not self._plain_terminal:
             if self._console_state == ConsoleState.STATUS:
-                cursor_position = "\033[A"
+                cursor_position = Cursor.UP()
             else:
                 cursor_position = ""
-            print("%-80s" % (cursor_position + Style.BRIGHT + message + Style.RESET_ALL,))
+            self._uprint(cursor_position, Style.BRIGHT, "%-80s"% message, Style.RESET_ALL)
             self._console_state = ConsoleState.STATUS
         else:
-            print(Style.BRIGHT + message + Style.RESET_ALL)
+            self._uprint(Style.BRIGHT, message, Style.RESET_ALL)
+
+    def _uprint(self, *args):
+        args = str().join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
+        print(args)
 
     def _print(self, *args, **kwargs):
+        # filter out any args that are escape codes and turn them into a single element tuple
+        args = str().join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
+        args = [args]
+
+        # continue to do the rest of the stupid stuff
         encoded_args = []
         if sys.version_info[0] >= 3:
             string_type = str
@@ -348,7 +358,7 @@ class ConsoleApplication(object):
             self._print(text)
         else:
             color = Fore.RED if level == 'error' else Fore.YELLOW
-            self._print(color + Style.BRIGHT + text + Style.RESET_ALL)
+            self._print(color, Style.BRIGHT, text, Style.RESET_ALL)
 
 
 def find_device(type):
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -132,7 +132,7 @@ class ConsoleApplication(object):
         self._exit_status = None
         self._console_state = ConsoleState.EMPTY
         self._have_terminal = sys.stdin.isatty() and sys.stdout.isatty() and not os.environ.get("TERM", '') == "dumb"
-        self._plain_terminal = os.environ.get("TERM", '').lower() == "none"
+        self._plain_terminal = os.environ.get("TERM", "").lower() == "none"
         self._quiet = False
         if sum(map(lambda v: int(v is not None), (self._device_id, self._device_type, self._host))) > 1:
             parser.error("Only one of -D, -U, -R, and -H may be specified")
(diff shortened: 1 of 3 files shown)


oleavr:
Timed out.

arizvisa:
What was the word on this?

oleavr:
Review issues need to be addressed. I'm overwhelmed so I didn't have the bandwidth for it myself, so if you're interested in fixing the remaining issues I'd be happy to merge it. Cheers!

arizvisa:
What were the review issues? I didn't see a review of the PR?

oleavr:
@arizvisa Ahh I'm so sorry, I somehow failed to submit them and thought I did. 🤦‍♂️ Reopening.

frida_tools/application.py
```diff
@@ -132,6 +132,7 @@ def store_target(option, opt_str, target_value, parser, target_type, *args, **kw
         self._exit_status = None
         self._console_state = ConsoleState.EMPTY
         self._have_terminal = sys.stdin.isatty() and sys.stdout.isatty() and not os.environ.get("TERM", '') == "dumb"
+        self._plain_terminal = os.environ.get("TERM", '').lower() == "none"
```
oleavr: Default value should be `""`, not `''`. (The convention used is double-quoted strings for strings, and single-quoted for enum values. In hindsight kinda weird, but better stick to it for consistency.)

frida_tools/application.py
```diff
@@ -308,25 +309,34 @@ def _on_device_lost(self):
...
+    def _uprint(self, *args):
+        args = str().join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
+        print(args)
 
     def _print(self, *args, **kwargs):
+        # filter out any args that are escape codes and turn them into a single element tuple
+        args = str().join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
+        args = [args]
+
+        # continue to do the rest of the stupid stuff
```
oleavr: This comment does not convey anything that the code doesn't, so it should be omitted.

frida_tools/application.py
```diff
@@ -308,25 +309,34 @@ def _on_device_lost(self):
...
                 cursor_position = ""
-            print("%-80s" % (cursor_position + Style.BRIGHT + message + Style.RESET_ALL,))
+            self._uprint(cursor_position, Style.BRIGHT, "%-80s"% message, Style.RESET_ALL)
             self._console_state = ConsoleState.STATUS
         else:
-            print(Style.BRIGHT + message + Style.RESET_ALL)
+            self._uprint(Style.BRIGHT, message, Style.RESET_ALL)
+
+    def _uprint(self, *args):
+        args = str().join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
```
oleavr: - `""` is slightly clearer than `str()`.
- Should use double quotes.

frida_tools/application.py
```diff
@@ -308,25 +309,34 @@ def _on_device_lost(self):
...
+        print(args)
 
     def _print(self, *args, **kwargs):
+        # filter out any args that are escape codes and turn them into a single element tuple
```
oleavr: Comments should use a capital letter at the start of each sentence, and each should end in a full stop. However, I'm not sure this comment is needed. But if it's needed we should instead add a helper function named appropriately.

frida_tools/application.py
```diff
@@ -308,25 +309,34 @@ def _on_device_lost(self):
...
+        args = str().join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
```
oleavr: Should use double-quotes here as well.

frida_tools/tracer.py
```diff
@@ -112,9 +112,9 @@ def on_trace_events(self, events):
                     indent = depth * "   | "
                     attributes = self._get_attributes(thread_id)
                     if thread_id != self._last_event_tid:
-                        self._print("%s           /* TID 0x%x */%s" % (attributes, thread_id, Style.RESET_ALL))
+                        self._print(attributes, '           ', "/* TID %#x */" % thread_id, Style.RESET_ALL)
                         self._last_event_tid = thread_id
-                    self._print("%6d ms  %s%s%s%s" % (timestamp, attributes, indent, message, no_attributes))
```
oleavr: Strings should be double-quoted.

frida_tools/tracer.py
```diff
@@ -112,9 +112,9 @@ def on_trace_events(self, events):
                     indent = depth * "   | "
                     attributes = self._get_attributes(thread_id)
                     if thread_id != self._last_event_tid:
-                        self._print("%s           /* TID 0x%x */%s" % (attributes, thread_id, Style.RESET_ALL))
```
oleavr: Strings should be double-quoted.

oleavr:
Bump :-)

oleavr:
Timed out.

arizvisa:
Lol. Whut? It's been closed within only 2 weeks? It took a whole two months to respond to the opening of the PR..

arizvisa:
And then 4 months to actually figure out what was wrong with it which turned out to be just code inconsistency issues? This is pretty lol. Is this really an opensource project?

As per my comment on #7 
```
me: ... writing code sucks, but if i submit a PR will you actually merge it?
you: Yes if it's a PR adding an option to disable it.
```

oleavr:
The initial review delay is what happens when an OSS maintainer is overloaded, it definitely wasn't intentional. As for the follow-up, I've done some triage passes where I closed potentially stale PRs; but I'll obviously reopen them if the submitter gets back and addresses the remaining changes.

> Is this really an opensource project?

It's an OSS project where the author cares deeply about the code being clean and consistent. Like a lot of other projects it's also one that has a maintainer spread too thin. This is unfortunate and I'm sorry for the frustration it caused you.

frida_tools/application.py
```diff
@@ -308,25 +309,33 @@ def _on_device_lost(self):
...
-            print(Style.BRIGHT + message + Style.RESET_ALL)
+            self._uprint(Style.BRIGHT, message, Style.RESET_ALL)
+
+    def _uprint(self, *args):
+        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
+        print(args)
 
     def _print(self, *args, **kwargs):
+        # Filter out any args that are escape codes and turn them into a single element tuple
+        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
```
oleavr:
```suggestion
        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith("\033"))
```

frida_tools/application.py
```diff
@@ -308,25 +309,33 @@ def _on_device_lost(self):
...
                 cursor_position = ""
-            print("%-80s" % (cursor_position + Style.BRIGHT + message + Style.RESET_ALL,))
+            self._uprint(cursor_position, Style.BRIGHT, "%-80s"% message, Style.RESET_ALL)
             self._console_state = ConsoleState.STATUS
         else:
-            print(Style.BRIGHT + message + Style.RESET_ALL)
+            self._uprint(Style.BRIGHT, message, Style.RESET_ALL)
+
+    def _uprint(self, *args):
+        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith('\033'))
```
oleavr:
```suggestion
        args = "".join(arg for arg in args if not self._plain_terminal or not arg.startswith("\033"))
```

frida_tools/tracer.py
```diff
@@ -112,9 +112,9 @@ def on_trace_events(self, events):
...
+                        self._print(attributes, "           ", "/* TID %#x */" % thread_id, Style.RESET_ALL)
                         self._last_event_tid = thread_id
-                    self._print("%6d ms  %s%s%s%s" % (timestamp, attributes, indent, message, no_attributes))
+                    self._print("%6d ms"% timestamp, "  ", attributes, indent, message, no_attributes)
```
```suggestion
                    self._print("%6d ms" % timestamp, "  ", attributes, indent, message, no_attributes)
```

frida_tools/application.py
```diff
@@ -308,25 +309,33 @@ def _on_device_lost(self):
...
     def _update_status(self, message):
-        if self._have_terminal:
+        if self._have_terminal and not self._plain_terminal:
             if self._console_state == ConsoleState.STATUS:
-                cursor_position = "\033[A"
+                cursor_position = Cursor.UP()
             else:
                 cursor_position = ""
-            print("%-80s" % (cursor_position + Style.BRIGHT + message + Style.RESET_ALL,))
+            self._uprint(cursor_position, Style.BRIGHT, "%-80s"% message, Style.RESET_ALL)
```
oleavr:
```suggestion
            self._uprint(cursor_position, Style.BRIGHT, "%-80s" % message, Style.RESET_ALL)
```

frida_tools/application.py
```diff
@@ -308,25 +309,33 @@ def _on_device_lost(self):
...
         message = reason[0].upper() + reason[1:].replace("-", " ")
-        self._print(Fore.RED + Style.BRIGHT + message + Style.RESET_ALL)
+        self._print(Fore.RED, Style.BRIGHT, message, Style.RESET_ALL)
         self._exit(1)
 
     def _clear_status(self):
-        if self._console_state == ConsoleState.STATUS:
-            print("\033[A" + (80 * " "))
+        if not self._plain_terminal and self._console_state == ConsoleState.STATUS:
+            self._uprint(Cursor.UP(), (80 * " "))
```
oleavr:
```suggestion
            self._uprint(Cursor.UP(), 80 * " ")
```

frida_tools/application.py
```diff
@@ -16,7 +16,7 @@
     import msvcrt
 
 import colorama
-from colorama import Fore, Style
+from colorama import Fore, Style, Cursor
```
oleavr:
```suggestion
from colorama import Cursor, Fore, Style
```

oleavr: (Imports should be alphabetically sorted.)

arizvisa:
the types of changes suggested in yourreview (that resulted in this being dropped w/o a review and then reopened+closed)  seemed to only be spacing and other seriously minor formatting issues, it'd probably save you a ton of time to either write up a CONTRIBUTING.md document with your personal preferences for committers, or better yet to include a lint config to hook commits with so that way you don't even need to take the time to manually go through and identify all the spacing issues in each commit.

definitely can be automated as i'm sure you'll agree that your time is better spent fixing bugs and guiding development.

arizvisa:
> ⚔️

(literally committed your suggestions since you essentially did all the work)

oleavr:
Thanks! ❤️

Totally agree with you that all of that would be nice, but this is just one of many repos, most of them very different and each requiring quite some effort to implement the same. However, I'm still going to have to review each PR and catch what they missed – because even when there's a guide, it's easy to overlook or misunderstand something. And no automated tool is going to make a call on where a blank line would improve readability, e.g. by separating logical operations part of a function. So in the end, I'd rather put in a few moments of extra time on the few PRs that we get, than put in a lot of up-front effort to optimize work that's very infrequent. Once that frequency increases, however, I will definitely prioritize optimizing that process. But in the meantime, I'd rather spend that time on more fruitful efforts.

This PR was however a PR that fell between two chairs due to me juggling too many things, so I hope you can forgive me for making this a very frustrating experience for you. Really appreciate your perseverance in getting this landed 🤗

oleavr:
FYI this caused a regression fixed in 3e722bdf50a0c0308de5824c61f8c67447c40b95. Ended up reverting the `_print()` changes and using colorama's “strip” feature instead. Let me know if you find any corner-cases where this doesn't work.

arizvisa:
thx for taking care of it.

PR #11, opened 2019-01-26
Upgrade python-prompt-toolkit to version 2.

leonjza:


setup.py
```diff
@@ -14,7 +14,7 @@
     install_requires=[
         "colorama >= 0.2.7, < 1.0.0",
         "frida >= 12.1.0, < 13.0.0",
-        "prompt-toolkit >= 0.57, < 2.0.0",
+        "prompt-toolkit >= 2.0.0, < 2.1.0",
```
oleavr: Shouldn't this be:

```suggestion
        "prompt-toolkit >= 2.0.0, < 3.0.0",
```

PR #18, opened 2019-05-11
[repl][completion-epc] fixed imports order and added missing import sys

otadmor:
sys is used on the obselete completion-epc-server instead of the newer completion-epc-client.


oleavr:
Thanks!

PR #19, opened 2019-05-23
convert any paths into absolute which file monitor expects

pushad:
On windows I've discovered that frida-tools (CLI) does not auto reload when using relative paths.
Making use of the os.path.abspath we convert the relative path into an absolute path.

### DIFF
--- a/frida_tools/repl.py
+++ b/frida_tools/repl.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals, print_function
 import os
 import signal
 import threading
+from os.path import abspath
 
 
 def main():
@@ -172,7 +173,7 @@ def main():
             self._unmonitor_script()
 
             if self._user_script is not None:
-                monitor = frida.FileMonitor(self._user_script)
+                monitor = frida.FileMonitor(abspath(self._user_script))
                 monitor.on('change', self._on_change)
                 monitor.enable()
                 self._script_monitor = monitor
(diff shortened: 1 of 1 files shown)


frida_tools/repl.py
```diff
@@ -172,7 +173,7 @@ def _monitor_script(self):
             self._unmonitor_script()
 
             if self._user_script is not None:
-                monitor = frida.FileMonitor(self._user_script)
+                monitor = frida.FileMonitor(abspath(self._user_script))
```
oleavr: Would be good to do this in `_initialize` instead:

```Python
        def _initialize(self, parser, options, args):
            self._user_script = os.path.abspath(options.user_script)
```

(And we can use the `os` import that we already have, as the function is only used once and not in a performance-critical path.)

pushad: That makes sense. Thanks for the update.
Will modify and update when I can!

pushad: Applied suggested changes.

frida_tools/repl.py
```diff
@@ -67,7 +66,7 @@ def _add_options(self, parser):
             parser.add_option("-o", "--output", help="output to log file", dest="logfile", default=None)
 
         def _initialize(self, parser, options, args):
-            self._user_script = options.user_script
+            self._user_script = os.path.abspath(options.user_script)
```
oleavr: I guess this will crash if `options.user_script` is `None` because `-l` wasn't specified?

oleavr:
Fixed in 3ca7f468a7158f3f307132a55b31a14e5889a6ee.

PR #21, opened 2019-06-26
Fix parameter error in OutputFile constructor

ShawnAlexander:
The `frida-trace` command was throwing errors about an invalid output filename. The `filename` parameter in `tracer.py` was qualified with a `self` reference where it shouldn't be.


oleavr:
Oops, nice catch! Thanks!

PR #26, opened 2019-11-14
-i/-x OPTIONS SUPPORT PATTERN FORM '<MODULE-REGEX>!<FUNCTION-REGEX>'

mosherubin:
Currently the -i (IncludeFunction) and -x (ExcludeFunction) command
line options take a function regex pattern, but will match functions
in *ALL* modules in the process.  There is no easy way to request a
function regex pattern in a *single* module.

This commit enhances the -i and -x function options to support patterns
that specify both a module *and* a function regex.

Examples:

`-i '*free*'`
        Matches all functions with 'free' in its name in ALL modules
        Equivalent to `exports:*!*free*`

`-i 'msvcrt.dll!*cpy*'`
        Matches all functions with 'cpy' only in msvcrt.dll
        Equivalent to `exports:msvcrt.dll!*cpy*`

`-i 'msvcrt.dll!'`
        Equivalent to `exports:msvcrt.dll:*`

`-i '!*cpy*'`
        Equivalent to `exports:*!*cpy*`

PR #27, opened 2019-11-26
New feature: initialize <state> variable with JS scripts before starting trace session

mosherubin:
This feature implementation enables initializing the global <state> variable with user-defined functions and variables before the trace session starts.  One major use of this feature is to share handler functions across developers, resulting in reusable code.

The feature consists of the -k / --init-session command line options followed by a path to a valid Javascript file.  This file will be executed via eval() before the tracing session begins.

See the Git commit message for caveats.

### DIFF
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -347,20 +347,16 @@ function excludeModule(pattern, workingSet) {
     return workingSet;
 }
 
-function parseExportsFunctionPattern (pattern) {
-    var m;
-    var f;
-    var res = pattern.split ('!');
-    
-    if (res.length == 1) {
+function parseExportsFunctionPattern(pattern) {
+    var res = pattern.split('!');
+    var m, f;
+    if (res.length === 1) {
         m = '*';
         f = res[0];
+    } else {
+        m = (res[0] === '') ? '*' : res[0];
+        f = (res[1] === '') ? '*' : res[1];
     }
-    else {
-        m = (res[0] == '' ? '*' : res[0]);
-        f = (res[1] == '' ? '*' : res[1]);
-    }
-    
     return {
         module: m,
         function: f
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -53,9 +53,9 @@ def main():
             parser.add_option("-s", "--include-debug-symbol", help="include DEBUG_SYMBOL", metavar="DEBUG_SYMBOL",
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
-            parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
-            parser.add_option("-S", "--init-session", help="Path to JavaScript file used to initialize the session", metavar="PATH",
-                    type='string', action='append')
+            parser.add_option("-d", "--decorate", help="add module name to generated onEnter log statement", action='store_true', default=False)
+            parser.add_option("-S", "--init-session", help="path to JavaScript file used to initialize the session", metavar="PATH",
+                    type='string', action='append', default=[])
             parser.add_option("-o", "--output", help="dump messages to file", metavar="OUTPUT", type='string')
             self._profile_builder = pb
 
@@ -161,7 +161,6 @@ class TracerProfileBuilder(object):
 
     def __init__(self):
         self._spec = []
-        self._init_session_paths = []
 
     def include_modules(self, *module_name_globs):
         for m in module_name_globs:
@@ -212,11 +211,6 @@ class TracerProfileBuilder(object):
     def include_debug_symbol(self, *function_name_globs):
         for f in function_name_globs:
             self._spec.append(('include', 'debug_symbol', f))
-        return self
-
-    def init_session_paths(self, *function_name_globs):
-        for f in function_name_globs:
-            self._init_session_paths.append(f)
         return self
 
     def build(self):
@@ -528,8 +522,7 @@ class Tracer(object):
         self._repository = repository
         self._profile = profile
         self._script = None
-        self._init_session_paths = init_session
-        self._init_session_code = ''
+        self._init_session_paths = init_session_paths
         self._log_handler = log_handler
 
     def start_trace(self, session, runtime, ui):
@@ -555,6 +548,9 @@ class Tracer(object):
         ui.on_trace_progress('resolve')
         working_set = self._profile.resolve(session, runtime, log_handler=self._log_handler)
         ui.on_trace_progress('instrument')
+        print ('----- _create_trace_script()  ------------------------------')  #debug
+        print (self._create_trace_script()) #debug
+        print ('------------------------------------------------------------')  #debug
         self._script = session.create_script(name="tracer",
                                              source=self._create_trace_script(),
                                              runtime=runtime)
@@ -585,17 +581,11 @@ class Tracer(object):
 
     def _create_trace_script(self):
         init_session_code = ""
-        for path in init_session_paths:
+        for path in self._init_session_paths:
             with open(path, 'rt') as fo:
                 lines = fo.readlines()
 
-            lines = [line.rstrip() for line in lines]
-
-            # eval() does not like 0x0A, replace with "\u000a"
-            init_session_code += "\\u000a".join(lines)
-            
-            # Insert "\ua000" in case another init session follows
-            init_session_code += "\\u000a"
+            init_session_code += "\n".join(lines)
 
         return """'use strict';
 
@@ -606,10 +596,12 @@ var pending = [];
 var timer = null;
 
 state.newline = String.fromCharCode(10);
-
-initializeSession();
+state.bar = "I am BAR!";
 
 rpc.exports = {
+    init: function (stage, parameters) {
+        %(init_code)s
+    },
     dispose: function () {
         flush();
     },
@@ -705,10 +697,6 @@ function parseHandler(target) {
         });
         return {};
     }
-}
-
-function initializeSession() {
-    eval("%(init_code)s");
 }
 """ % {"init_code": init_session_code}
 
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -54,6 +54,8 @@ def main():
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
             parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
+            parser.add_option("-k", "--init-session", help="Path to Javascript file used to initialize the session", metavar="PATH",
+                    type='string', action='callback', callback=process_builder_arg, callback_args=(pb.init_session_paths,))
             parser.add_option("-o", "--output", help="dump messages to file", metavar="OUTPUT", type='string')
             self._profile_builder = pb
 
@@ -75,7 +77,7 @@ def main():
         def _start(self):
             if self._output_path is not None:
                 self._output = OutputFile(self._output_path)
-            self._tracer = Tracer(self._reactor, FileRepository(self._reactor, self._decorate), self._profile, log_handler=self._log)
+            self._tracer = Tracer(self._reactor, FileRepository(self._reactor, self._decorate), self._profile, self._profile_builder._init_session_paths, log_handler=self._log)
             try:
                 self._targets = self._tracer.start_trace(self._session, self._runtime, self)
             except Exception as e:
@@ -158,6 +160,7 @@ class TracerProfileBuilder(object):
 
     def __init__(self):
         self._spec = []
+        self._init_session_paths = []
 
     def include_modules(self, *module_name_globs):
         for m in module_name_globs:
@@ -208,6 +211,11 @@ class TracerProfileBuilder(object):
     def include_debug_symbol(self, *function_name_globs):
         for f in function_name_globs:
             self._spec.append(('include', 'debug_symbol', f))
+        return self
+
+    def init_session_paths(self, *function_name_globs):
+        for f in function_name_globs:
+            self._init_session_paths.append(f)
         return self
 
     def build(self):
@@ -530,11 +538,13 @@ function debugSymbolFromAddress(address) {
 
 
 class Tracer(object):
-    def __init__(self, reactor, repository, profile, log_handler=None):
+    def __init__(self, reactor, repository, profile, init_session, log_handler=None):
         self._reactor = reactor
         self._repository = repository
         self._profile = profile
         self._script = None
+        self._init_session_paths = init_session
+        self._init_session_code = ''
         self._log_handler = log_handler
 
     def start_trace(self, session, runtime, ui):
@@ -589,6 +599,21 @@ class Tracer(object):
             self._script = None
 
     def _create_trace_script(self):
+        # Extract all init session files into a single string,
+        # substitute newlines with '\u000a'
+        self._init_session_code = ''
+        for path in self._init_session_paths:
+            # Append contents of <path> to <self._init_session_code>
+            fo = open(path, "rt")
+            line = fo.readlines()
+            fo.close()
+
+            for i in range (len(line)):
+                line [i] = line [i].rstrip ()
+
+            self._init_session_code += '\\u000a'.join(line)
+            self._init_session_code += '\\u000a'
+
         return """'use strict';
 
 var started = Date.now();
@@ -596,6 +621,13 @@ var handlers = {};
 var state = {};
 var pending = [];
 var timer = null;
+
+// Create namespaces for user functions ('F') and data ('D'), pre-populate if relevant
+state.F = {}
+state.D = {}
+state.D.NEWLINE = String.fromCharCode (10);
+
+initializeSession();
 
 rpc.exports = {
     dispose: function () {
@@ -694,7 +726,11 @@ function parseHandler(target) {
         return {};
     }
 }
-"""
+
+function initializeSession() {
+    eval ("%(init_code)s");
+}
+""" % {"init_code": self._init_session_code}
 
     def _process_message(self, message, data, ui):
         handled = False
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -161,6 +161,7 @@ class TracerProfileBuilder(object):
 
     def __init__(self):
         self._spec = []
+        self._init_session_paths = []
 
     def include_modules(self, *module_name_globs):
         for m in module_name_globs:
@@ -211,6 +212,11 @@ class TracerProfileBuilder(object):
     def include_debug_symbol(self, *function_name_globs):
         for f in function_name_globs:
             self._spec.append(('include', 'debug_symbol', f))
+        return self
+
+    def init_session_paths(self, *function_name_globs):
+        for f in function_name_globs:
+            self._init_session_paths.append(f)
         return self
 
     def build(self):
@@ -522,6 +528,8 @@ class Tracer(object):
         self._repository = repository
         self._profile = profile
         self._script = None
+        self._init_session_paths = init_session
+        self._init_session_code = ''
         self._log_handler = log_handler
 
     def start_trace(self, session, runtime, ui):
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -625,7 +625,7 @@ var timer = null;
 // Create namespaces for user functions ('F') and data ('D'), pre-populate if relevant
 state.F = {}
 state.D = {}
-state.D.NEWLINE = String.fromCharCode (10);
+state.D.NEWLINE = String.fromCharCode(10);
 
 initializeSession();
 
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -53,6 +53,7 @@ def main():
             parser.add_option("-s", "--include-debug-symbol", help="include DEBUG_SYMBOL", metavar="DEBUG_SYMBOL",
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
+            parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
             parser.add_option("-o", "--output", help="dump messages to file", metavar="OUTPUT", type='string')
             self._profile_builder = pb
 
@@ -64,6 +65,7 @@ def main():
             self._targets = None
             self._profile = self._profile_builder.build()
             self._quiet = options.quiet
+            self._decorate = options.decorate
             self._output = None
             self._output_path = options.output
 
@@ -73,7 +75,7 @@ def main():
         def _start(self):
             if self._output_path is not None:
                 self._output = OutputFile(self._output_path)
-            self._tracer = Tracer(self._reactor, FileRepository(self._reactor), self._profile, log_handler=self._log)
+            self._tracer = Tracer(self._reactor, FileRepository(self._reactor, self._decorate), self._profile, log_handler=self._log)
             try:
                 self._targets = self._tracer.start_trace(self._session, self._runtime, self)
             except Exception as e:
@@ -366,21 +368,37 @@ function parseExportsFunctionPattern (pattern) {
 }
 
 function includeFunction(pattern, workingSet) {
-    var obj = parseExportsFunctionPattern (pattern);
-    
-    moduleResolver().enumerateMatches('exports:' + obj['module'] + '!' + obj['function']).forEach(function (m) {
+    var parseResult = parseExportsFunctionPattern(pattern);
+    moduleResolver().enumerateMatches('exports:' + parseResult.module + '!' + parseResult.function).forEach(function (m) {
         workingSet[m.address.toString()] = moduleExportFromMatch(m);
     });
     return workingSet;
 }
 
 function excludeFunction(pattern, workingSet) {
-    var obj = parseExportsFunctionPattern (pattern);
-    
-    moduleResolver().enumerateMatches('exports:' + obj['module'] + '!' + obj['function']).forEach(function (m) {
+    var parseResult = parseExportsFunctionPattern(pattern);
+    moduleResolver().enumerateMatches('exports:' + parseResult.module + '!' + parseResult.function).forEach(function (m) {
         delete workingSet[m.address.toString()];
     });
     return workingSet;
+}
+
+function parseExportsFunctionPattern(pattern) {
+    var res = pattern.split('!');
+
+    var m, f;
+    if (res.length === 1) {
+        m = '*';
+        f = res[0];
+    } else {
+        m = (res[0] === '') ? '*' : res[0];
+        f = (res[1] === '') ? '*' : res[1];
+    }
+
+    return {
+        module: m,
+        function: f
+    };
 }
 
 function includeRelativeFunction(func, workingSet) {
@@ -704,6 +722,7 @@ class Repository(object):
         self._on_create_callback = None
         self._on_load_callback = None
         self._on_update_callback = None
+        self._decorate = False
 
     def ensure_handler(self, function):
         raise NotImplementedError("not implemented")
@@ -732,7 +751,7 @@ class Repository(object):
         if self._on_update_callback is not None:
             self._on_update_callback(function, handler, source)
 
-    def _create_stub_handler(self, function):
+    def _create_stub_handler(self, function, decorate):
         if isinstance(function, ObjCMethod):
             display_name = function.display_name()
             state = {"index": 2}
@@ -796,16 +815,22 @@ class Repository(object):
                 except Exception as e:
                     pass
 
+            if decorate:
+                module_string = " [%s]" %  function.module.name
+            else:
+                module_string = ""
+
             if len(args) == 0:
-                log_str = "'%(name)s()'" % { "name": function.name }
+                log_str = "'%(name)s()%(module_string)s'" % { "name": function.name, "module_string" : module_string }
             else:
                 indent_outer = "    "
                 indent_inner = "      "
-                log_str = "'%(name)s(' +\n%(indent_inner)s%(args)s\n%(indent_outer)s')'" % {
+                log_str = "'%(name)s(' +\n%(indent_inner)s%(args)s\n%(indent_outer)s')%(module_string)s'" % {
                     "name": function.name,
                     "args": ("\n" + indent_inner).join(args),
                     "indent_outer": indent_outer,
-                    "indent_inner": indent_inner
+                    "indent_inner": indent_inner,
+                    "module_string": module_string
                 }
 
         return """\
@@ -858,7 +883,7 @@ class MemoryRepository(Repository):
     def ensure_handler(self, function):
         handler = self._handlers.get(function)
         if handler is None:
-            handler = self._create_stub_handler(function)
+            handler = self._create_stub_handler(function, False)
             self._handlers[function] = handler
             self._notify_create(function, handler, "memory")
         else:
@@ -867,7 +892,7 @@ class MemoryRepository(Repository):
 
 
 class FileRepository(Repository):
-    def __init__(self, reactor):
+    def __init__(self, reactor, decorate):
         super(FileRepository, self).__init__()
         self._reactor = reactor
         self._handler_by_address = {}
@@ -876,6 +901,7 @@ class FileRepository(Repository):
         self._last_change_id = 0
         self._repo_dir = os.path.join(os.getcwd(), "__handlers__")
         self._repo_monitors = {}
+        self._decorate = decorate
 
     def ensure_handler(self, function):
         entry = self._handler_by_address.get(function.absolute_address)
@@ -902,7 +928,7 @@ class FileRepository(Repository):
                 break
 
         if handler is None:
-            handler = self._create_stub_handler(function)
+            handler = self._create_stub_handler(function, self._decorate)
             handler_file = handler_files_to_try[0]
             handler_dir = os.path.dirname(handler_file)
             if not os.path.isdir(handler_dir):
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -345,15 +345,39 @@ function excludeModule(pattern, workingSet) {
     return workingSet;
 }
 
+function parseExportsFunctionPattern (pattern) {
+    var m;
+    var f;
+    var res = pattern.split ('!');
+    
+    if (res.length == 1) {
+        m = '*';
+        f = res[0];
+    }
+    else {
+        m = (res[0] == '' ? '*' : res[0]);
+        f = (res[1] == '' ? '*' : res[1]);
+    }
+    
+    return {
+        module: m,
+        function: f
+    };
+}
+
 function includeFunction(pattern, workingSet) {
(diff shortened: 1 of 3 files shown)


mosherubin:
Hi Oleavr,

I see it says above "_Add more commits by pushing to the init_session_impl branch on mosherubin/frida-tools_".  I have two more features I am actively using, and would like to return them to Frida master.  Do I need to open two additional branches off of master, or, as per the text above, can I just add more commits and pull requests to my branch "init_session_impl"?

oleavr:
Hey! Thanks for doing this! ❤️ (And sorry for the delay, I've been heads down in unrelated code.) You will have to create each such branch based on upstream `master`, otherwise the PRs will look very confusing.

frida_tools/tracer.py
```diff
@@ -54,6 +54,8 @@ def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
             parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
+            parser.add_option("-k", "--init-session", help="Path to Javascript file used to initialize the session", metavar="PATH",
```
```suggestion
            parser.add_option("-S", "--init-session", help="Path to JavaScript file used to initialize the session", metavar="PATH",
```

Perhaps we can use `-S`? Also tweaked casing of `JavaScript`.

frida_tools/tracer.py
```diff
@@ -573,6 +603,21 @@ def stop(self):
...
 
     def _create_trace_script(self):
+        # Extract all init session files into a single string,
+        # substitute newlines with '\u000a'
+        self._init_session_code = ''
+        for path in self._init_session_paths:
+            # Append contents of <path> to <self._init_session_code>
+            fo = open(path, "rt")
+            line = fo.readlines()
+            fo.close()
```
oleavr: Better to use `with` here so there's no need to call `.close()`.

frida_tools/tracer.py
```diff
@@ -573,6 +603,21 @@ def stop(self):
             self._script = None
 
     def _create_trace_script(self):
+        # Extract all init session files into a single string,
+        # substitute newlines with '\u000a'
```
oleavr: As outlined in the [style guide](https://github.com/frida/frida/blob/master/CONTRIBUTING.md), this comment should probably be omitted in favor of naming variables, or a helper function named appropriately.

frida_tools/tracer.py
```diff
@@ -573,6 +603,21 @@ def stop(self):
...
+            # Append contents of <path> to <self._init_session_code>
+            fo = open(path, "rt")
+            line = fo.readlines()
+            fo.close()
+
+            for i in range (len(line)):
+                line [i] = line [i].rstrip ()
+
+            self._init_session_code += '\\u000a'.join(line)
+            self._init_session_code += '\\u000a'
```
oleavr: Wouldn't `\n` have the same effect? Also it would be good to use double-quoted strings here, as these are not enum values.

frida_tools/tracer.py
```diff
@@ -581,6 +626,13 @@ def _create_trace_script(self):
 var pending = [];
 var timer = null;
 
+// Create namespaces for user functions ('F') and data ('D'), pre-populate if relevant
+state.F = {}
+state.D = {}
+state.D.NEWLINE = String.fromCharCode(10);
```
oleavr: How about naming these `$userFunctions` and `$userData`, assuming they're "special"? (In any case we shouldn't violate the style by using PascalCase for properties – properties should be camelCase.) Are these meant to be accessed by the user or are they internals?

frida_tools/tracer.py
```diff
@@ -678,7 +730,11 @@ def _create_trace_script(self):
         return {};
     }
 }
-"""
+
+function initializeSession() {
+    eval ("%(init_code)s");
```
```suggestion
    eval("%(init_code)s");
```

frida_tools/tracer.py
```diff
@@ -573,6 +603,21 @@ def stop(self):
...
+        # substitute newlines with '\u000a'
+        self._init_session_code = ''
+        for path in self._init_session_paths:
+            # Append contents of <path> to <self._init_session_code>
+            fo = open(path, "rt")
+            line = fo.readlines()
+            fo.close()
+
+            for i in range (len(line)):
+                line [i] = line [i].rstrip ()
```
oleavr: Extra spaces in function calls. (See style guide mentioned earlier.)

But how about using list comprehension here? E.g. `lines = [line.rstrip() for line in lines]`
(And renaming to `lines`, as it's more than one line.)

frida_tools/tracer.py
```diff
@@ -347,6 +355,26 @@ def _create_resolver_script(self):
...
+    else {
+        m = (res[0] == '' ? '*' : res[0]);
+        f = (res[1] == '' ? '*' : res[1]);
+    }
+    
+    return {
+        module: m,
+        function: f
+    };
+}
```
oleavr: This looks like unrelated code?

frida_tools/tracer.py
```diff
@@ -54,6 +54,8 @@ def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
             parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
+            parser.add_option("-k", "--init-session", help="Path to Javascript file used to initialize the session", metavar="PATH",
+                    type='string', action='callback', callback=process_builder_arg, callback_args=(pb.init_session_paths,))
```
oleavr: Does this belong on the builder though? It looks like the logic is outside it, so we probably shouldn't store the state on it.

frida_tools/tracer.py
mosherubin: I hadn't noticed this style guide, although you pointed several issues out to me.  Thanks for the pointer!

frida_tools/tracer.py
mosherubin: There is a reason for this madness. :-)

I discovered a very important, albeit troubling, behaviour of JavaScript's `eval` function: newlines will cause the eval() call to fail!  This includes two forms of newlines:

(a) **The occurrence of the two adjacent characters backslash ('\') and 'n'**.  Upon reading these within a string, eval() assumes we have come to the end of the line, so the string it was parsing will be unterminated.

(b) **The occurrence of the binary value 0x0A in the input**.  Typically this is due to the developer having pressed ENTER between code lines.  Encountering this causes eval() to throw an error message.

It took me a while to pin this down and solve it.  The solution to both newline problems is:

(a) **The developer must never insert '\n' within the evaluated JS code**.  Instead, s/he should use the String.fromCharCode(10) function call, concatenating this string wherever '\n' is needed.  Although this is equivalent to '\n' at run-time, eval() does not see the '\n' sequence at all.

(b) **Replace the 0x0A between code lines with '\u000a'**.  It turns out eval() accepts this as a valid line delimiter, and does **not** throw an error.  The developer does not have to worry about this -- tracer.py does it for him automatically.

It's interesting to note that *all* eval() examples on the web are toy examples: they always consist of one or more JS statements *on a single line*.  Now we know why they shy away from multiple lines -- because of newlines.  Fortunately, frida-trace can benefit from the workaround.

frida_tools/tracer.py
mosherubin: The are meant to be access by the user.  So, for example, a developer might create a function called `dump_function_parameters` (a real example we have done).  The init-session JS script defines this function, storing it in `state`.  I intentionally wanted to create the shortest possible name (`state.F`) to reduce noise in handler functions that call this function.  The function call:

`state.F.dump_function_parameters (parms)`

is shorter than:

`state.userFunctions.dump_function_parameters (parms)`.

Having said that, I actually like your suggestion, for clarity's sake.  If you have no problem with the possible verbosity, I think `userFunctions` and `userData` are quite acceptable.  Awaiting your decision.

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: Done (good point!).

frida_tools/tracer.py
mosherubin: I have removed the comments from the beginning of the function, but feel it is crucial to document the two lines that insert '\u000a':

```
def _create_trace_script(self):
    self._init_session_code = ''
    for path in self._init_session_paths:
        # Append contents of <path> to <self._init_session_code>
        with open(path, 'rt') as fo
            line = fo.readlines()

        for i in range (len(line)):
            line [i] = line [i].rstrip ()

        # eval() does not like 0x0A, replace with '\u000a'
        self._init_session_code += '\\u000a'.join(line)

        # Insert '\ua000' in case another init session follows
        self._init_session_code += '\\u000a'

    return """'use strict';
```
It is too easy for another maintainer to restore the '\n' sequence, which will break the entire feature.  What do you think?

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: Good use of comprehensions (although I've been using Python for a decade, I still think in C <g>).

I've made the line -> lines change.

frida_tools/tracer.py
mosherubin: This is code added in a [previous pull request](https://github.com/frida/frida-tools/pull/26).  I don't know how it got into this PR, please ignore it.

frida_tools/tracer.py
mosherubin: I wanted to piggyback on the same logic used by the other command line options: store the --init-session paths during parsing, then process them when generating the JS trace code.  Can you suggest a better home for handling --init-session options?

mosherubin:
Regarding always creating a branch off of the upstream master, I thought I was very careful to do so.  I even compared the frida/frida-tools master and my own mosherubin/frida-tools master before branching.

Can you tell me if [the following web page](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) is the way to keep my fork in sync?

frida_tools/tracer.py
oleavr: Ah I see. Thanks for clarifying! Thinking about it more, we could also just evaluate those scripts in the global context (which I guess is already happening?), and have them be globally visible. That way any variables and functions are globally visible and conveniently accessible from the hooks. Would that work?

frida_tools/tracer.py
oleavr: Ahh! It would definitely be good to add a comment explaining this. :-)

frida_tools/tracer.py
oleavr: Yes definitely want to explain this bit. The `# Append contents` comment above can be removed though.

frida_tools/tracer.py
oleavr: I would store it on `self` like the other regular options.

frida_tools/tracer.py
oleavr: If you `git rebase origin/master`, it should clean up this PR.

frida_tools/tracer.py
```diff
@@ -514,11 +542,13 @@ def _create_resolver_script(self):
...
 
 class Tracer(object):
-    def __init__(self, reactor, repository, profile, log_handler=None):
+    def __init__(self, reactor, repository, profile, init_session, log_handler=None):
         self._reactor = reactor
         self._repository = repository
         self._profile = profile
         self._script = None
+        self._init_session_paths = init_session
+        self._init_session_code = ''
```
oleavr: It appears this variable is only used locally below, so there's no need to keep it around right?

frida_tools/tracer.py
mosherubin: The `#Append contents` should of course be deleted (don't know what I was smoking when I left it in :-))

frida_tools/tracer.py
oleavr: (And force-push after that: `git push -f`.)

frida_tools/tracer.py
mosherubin: I wanted to provide the developer with a convenient namespace for functions (`'state.F'`) and data (`'state.D'`).  This would prevent 'polluting' the global namespace.

To clarify this and other issues, I provide one of the `init-session` JS files I use in all my frida-trace work.  I now typically invoke frida-trace with:

`frida-trace -init-session /<path>/frida-library.general.js ...`.

In this way, whenever a handler's `onEnter` / `onLeave` is entered, I know that I can invoke any of the functions:

```
state.F.general_clone_args()
state.F.general_hexdump_lines(outlines, desc, nptr, len, indent_string)
state.F.general_hexdump(log, desc, nptr, len)
state.F.general_decode_bitflags(value, dict)
state.F.general_decode_enum(value, dict)
state.F.general_backtrace(log, desc, my_this)
state.F.process_function_parameters(outlines, funcname, my_args, parmdata, flags)
state.F.dump_function_parameters(log, funcname, my_args, parmdata, flags)
```

Here's the `init-session` file I referred to above (excuse the extra whitespace <g>).  I typically specify another `init-session` file on the command line for non-general functions (e.g., functions specific to a particular SO / DLL I'm working on):

```
// Define useful general-purpose Frida handler functions

//--------------------------------------------------------------------
// general_clone_args()
//--------------------------------------------------------------------

state.F.general_clone_args = function (args, num_args, my_this)
{
    if (my_this.args === undefined)
    {
        my_this.args = []

        for (var i=0; i<num_args; ++i)
        {
            my_this.args.push (args[i]);
        }
    }
};

//--------------------------------------------------------------------
// general_return_hexdump_lines()
//--------------------------------------------------------------------

state.F.general_hexdump_lines = function (outlines, desc, nptr, len, indent_string)
{
    if (nptr === undefined)
    {
        outlines.push (indent_string + 'INFO: (general_hexdump: ' + desc + ') nptr is UNDEFINED');
    }
    else if (nptr.isNull() === true)
    {
        outlines.push (indent_string + 'INFO: (general_hexdump: ' + desc + ') nptr is NULL');
    }
    else
    {
        outlines.push (desc);
        var s = hexdump (nptr, {length: len});
        outlines.push (indent_string + s);
    }
};

//--------------------------------------------------------------------
// general_hexdump()
//--------------------------------------------------------------------

state.F.general_hexdump = function (log, desc, nptr, len)
{
    var outlines = [];

    state.F.general_hexdump_lines (outlines, desc, nptr, len, '');
    log (outlines.join(NEWLINE));
};

//--------------------------------------------------------------------
// general_decode_bitflags()
//--------------------------------------------------------------------

state.F.general_decode_bitflags = function (value, dict)
{
    var s = '';

    for (var key in dict)
    {
        if (value & key)
        {
            if (s.length > 0)
            {
                s = s + ' | ';
            }

            s = s + dict[key];
        }
    }

    return s;
};

//--------------------------------------------------------------------
// general_decode_enum()
//--------------------------------------------------------------------

state.F.general_decode_enum = function (value, dict)
{
    for (var key in dict)
    {
        if (value == key)
        {
            return dict[key];
        }
    }

    return '<UNKNOWN>';
};

//--------------------------------------------------------------------
// general_backtrace()
//--------------------------------------------------------------------

state.F.general_backtrace_lines = function (outlines, desc, my_this)
{
    if (my_this !== undefined)
    {
        if (my_this.context !== undefined)
        {
            outlines.push (NEWLINE + '[*] backtrace (' + desc + '): ' + Thread.backtrace (my_this.context, Backtracer.ACCURATE).map(DebugSymbol.fromAddress).join(NEWLINE) + NEWLINE);
        }
        else
        {
            outlines.push (NEWLINE + '[*] backtrace (' + desc + '): No context found' + String.fromCharCode (10));
        }
    }
};

state.F.general_backtrace = function (log, desc, my_this)
{
    var outlines = [];
    state.F.general_backtrace_lines (outlines, desc, my_this);
    log (outlines.join (NEWLINE));
};

//--------------------------------------------------------------------
// process_function_parameters()
//
// This function accepts a function's name and its parameter metadata,
// performing the requested <flags> and writing any output to <log>.
//
// The format of <parmdata> is an array of objects, one object for each
// function parameter.
//
// Here is an example of invoking this function: <parmdata> example
//
//     process_function_parms (log,
//                             'ff_get_buffer',
//                             [
//                                 {name : 'avctx', type : 'AVCodecContext *'},
//                                 {name : 'frame', type : 'AVFrame *'},
//                                 {name : 'flags', type : 'int'}
//                             ],
//                             HEX_DUMP);
//
// Note:
//
//     (1) By default all parameter types are assumed to be 32-bit
//         (4 bytes).  We will enhance this function to support 64-bit
//         (8 byte) when the need arises.
//
//     (2) The number of asterisks in the parameter type string dictates
//         whether it is a non-ptr, a simple pointer, a pointer-to-a-pointer,
//         etc.  Thus, 'int' is a non-ptr, 'AVFrame *' is a simple pointer,
//         and 'undefined4 **' is a pointer-to-a-pointer.
//
//--------------------------------------------------------------------

state.D.GENERAL_HEX_DUMP     = 0x01;
state.D.GENERAL_BRIEF_OUTPUT = 0x02;

state.F.process_function_parameters = function (outlines, funcname, my_args, parmdata, flags)
{
    //----------------------
    //----------------------
    function ComputeNumberDereferencesInType (type)
    {
        var i;
        var count = 0;

        for (i=0; i<type.length; ++i)
        {
            if (type[i] == '*')
            {
                ++count;
            }
        }

        return count;
    }

    //----------------------
    //----------------------
    function process_data (outlines, native_pointer, pos, data, isBriefMode)
    {
        var i;
        var s = '';
        s = s + 'parameter ' + pos + ':';

        if ('type' in data)
        {
            s = s + ' ' + data.type;
        }

        if ('name' in data)
        {
            s = s + ' ' + data.name;
        }

        if (!isBriefMode)
        {
            // Caller did not specify brief output, add newline for readability
            s = NEWLINE + s;
        }
        outlines.push (s);

        var numDerefs = ComputeNumberDereferencesInType (data.type);
        var p = ptr(native_pointer);

        if (numDerefs === 0)
        {
            // Different output for non-pointer
            outlines.push ('\t[0] ' + p.toString());
            return;
        }

        for (i=0; i<numDerefs; ++i)
        {
            outlines.push ('\t[' + i + '] ' + p.toString());

            if (p.isNull() === true)
                break;

            if (i < numDerefs - 1)
                p = p.readPointer();
        }

        if (p.isNull() === false)
        {
            var t = '\t[' + numDerefs + '] ';
            var len;

            if ('len' in data)
                len = data.len;
            else
                len = 32;
            state.F.general_hexdump_lines (outlines, t, p, len, '\t\t');
        }
    }

    // Main portion of function
    var isBriefMode = (flags & state.D.GENERAL_BRIEF_OUTPUT);

    outlines.push (NEWLINE);
    outlines.push ('==========  BEGIN processing ' + funcname + ' (' + parmdata.length + ' parameters)' + '  ==========================');

    for (var i=0; i<parmdata.length; ++i)
    {
        var data = parmdata[i];
        process_data (outlines, my_args[i], i, data, isBriefMode);
    }

    var t = '==========  END   processing parameters: ' + funcname + '  ==========================' + String.fromCharCode (10);
    if (!isBriefMode)
        t = NEWLINE + t;
    outlines.push (t);
};

//--------------------------------------------------------------------
// dump_function_parameters()
//--------------------------------------------------------------------

state.F.dump_function_parameters = function (log, funcname, my_args, parmdata, flags)
{
    // Optional values
    flags = flags || 0;

    var outlines = [];
    state.F.process_function_parameters (outlines, funcname, my_args, parmdata, state.D.GENERAL_HEX_DUMP | flags);
    log (outlines.join (NEWLINE));
};
```

Note the use of `state.D` to store global constants used by the `state.F.dump_function_parameters` function.

So these functions **are** globally visible and accessible, but tucked under `state.F`.  Of course, the developer could decide to place `general_clone_args()` directly under `state` rather than `state.F`.  Do we want to 'help' the developer by creating `state.F` or to leave it up to him where he wants to place his functions?

frida_tools/tracer.py
mosherubin: Thanks for the pointer.  I performed:

- git stash
- git rebase origin/master
- git stash apply

I hope this clears things up when I push.

frida_tools/tracer.py
mosherubin: There can be multiple `init-session` options on the command line, so I need the callback method for storing all of the script paths.  Should `--init-session` be handled any differently than, say, `--include`?

frida_tools/tracer.py
mosherubin: Good catch!  I have removed `self._init_session_code`, making it local to `_create_trace_script()`:

```
    def _create_trace_script(self):
        var init_session_code = ''
        for path in self._init_session_paths:
            with open(path, 'rt') as fo
                lines = fo.readlines()

            lines = [line.rstrip() for line in lines]

            # eval() does not like 0x0A, replace with '\u000a'
            init_session_code += '\\u000a'.join(lines)
            
            # Insert '\ua000' in case another init session follows
            init_session_code += '\\u000a'

        return """'use strict';
...
""" % {"init_code": init_session_code}
```

frida_tools/tracer.py
oleavr: Well if it integrates with the builder then it should be stored there, but it doesn't so that's why I was thinking it should be stored outside. So if we store it outside, we should be able to use the `append` action.

frida_tools/tracer.py
oleavr: Cool! Remember to use double-quotes since it's considered a string and not an enum value.

frida_tools/tracer.py
oleavr: I think we want to leave it up to the developer, and make it very easy to add globally visible things, as frida-trace scripts are a bit "quick and dirty" in nature. :-)

By the way, as a side-note, function names and variable names should be camelCase in idiomatic JavaScript code. (Obviously users can do what they want but we should be careful so we use this style consistently in our code, to encourage following the most common style.)

mosherubin: You are right, the user should be free to store function and data object wherever he wants to.  I have therefore remove references to `state.F` and `state.D` in `tracer.py`.  Should he want to use a dedicated folder under `state`, he can define the folder(s) under state in an `--init-session` on the command line -- in this way, all handlers can assume the folders exist and will not need to test for their existence.

oleavr: Sounds good!

oleavr:
By the way, it doesn't look like any recent changes made it to the PR. Perhaps the push failed or went to the wrong branch?

frida_tools/tracer.py
mosherubin: It's not clear to me why some options (the regulars) go into `self` while others go into `builder`.  What is the concept of builder vs. non-builder?

Nonetheless, I have made the change, which is now much cleaner.  No need for callbacks to builder, the multiple -S parameters are appended to `self`.  Good point.

mosherubin:
Hi Oleavr,

OK, I pushed several fixes.  Regarding not having pushed my changes: I was under the obviously incorrect assumption that I should push my changes only after fixing all of your comments, rather than piecemeal.  If I now understand correctly, I can, and probably should, push whenever I fix an issue.  Is this correct?

frida_tools/tracer.py
mosherubin: Point well made and taken: if and when I get to document examples of init-session JS files, the examples should adhere to the project's JavaScript coding standards.

frida_tools/tracer.py
mosherubin: OK, done.

frida_tools/tracer.py
oleavr: Awesome, thanks!

The builder is currently only meant to construct the list of includes/excludes that specifies what to trace. But in retrospect, given its name Tracer**Profile**Builder, that name does suggest that it's supposed to specify a lot more than just the *what*. So it would make sense to rename it. 🤔 Ah well, let's keep that in mind for later.

frida_tools/tracer.py
oleavr: 🙌 Just to clarify, the camelCase style is quite pervasive across JavaScript ecosystems and projects, both web and Node.js – it's not something I'm advocating because I'm a fan of that particular style. :-) (If designing a language from scratch I would probably pick snake_case for variables and function/method names.)

oleavr:
> Can you tell me if the following web page is the way to keep my fork in sync?

That's definitely the safest way to do it, but personally I prefer to rebase instead of merging, to keep the feature-branch history readable. For larger features I try to separate changes into multiple commits, and use interactive rebasing (`git rebase -i`) to refine those commits as the work progresses. The rationale is to keep the history nice and readable for easier review, but also to avoid losing valuable history if upstream feels inclined to squash all of them into one commit at merge-time due to noisy commit history.

oleavr:
> If I now understand correctly, I can, and probably should, push whenever I fix an issue. Is this correct?

Ideally yes, but only if it's not a hassle for you. That way I might get a chance to look at that part of the PR, and resolve that conversation if it looks good, or provide early feedback if there's a misunderstanding.

frida_tools/tracer.py
```diff
@@ -54,6 +54,8 @@ def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
             parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
+            parser.add_option("-S", "--init-session", help="Path to JavaScript file used to initialize the session", metavar="PATH",
```
```suggestion
            parser.add_option("-S", "--init-session", help="path to JavaScript file used to initialize the session", metavar="PATH",
```

(For consistency with the other help messages. I totally missed this on the `--decorate` PR.)

frida_tools/tracer.py
```diff
@@ -158,6 +161,7 @@ class TracerProfileBuilder(object):
 
     def __init__(self):
         self._spec = []
+        self._init_session_paths = []
```
```suggestion
```

Left-over from the earlier approach.

frida_tools/tracer.py
```diff
@@ -210,6 +214,11 @@ def include_debug_symbol(self, *function_name_globs):
             self._spec.append(('include', 'debug_symbol', f))
         return self
 
+    def init_session_paths(self, *function_name_globs):
+        for f in function_name_globs:
+            self._init_session_paths.append(f)
+        return self
+
```
oleavr: Dead code left-over from the earlier approach.

frida_tools/tracer.py
```diff
@@ -68,14 +70,15 @@ def _initialize(self, parser, options, args):
             self._decorate = options.decorate
             self._output = None
             self._output_path = options.output
+            self._init_session_paths = options.init_session
```
oleavr: Come to think of it now, it would be good to actually read the files here. This is so the user gets quick feedback if a typo is made, instead of having to wait until we've attached to the target process. (We also do it early in our REPL if loading a user-defined script.)

frida_tools/tracer.py
```diff
@@ -514,11 +523,13 @@ def _create_resolver_script(self):
 
 
 class Tracer(object):
-    def __init__(self, reactor, repository, profile, log_handler=None):
+    def __init__(self, reactor, repository, profile, init_session_paths, log_handler=None):
         self._reactor = reactor
         self._repository = repository
         self._profile = profile
         self._script = None
+        self._init_session_paths = init_session
```
```suggestion
        self._init_session_paths = init_session_paths
```

frida_tools/tracer.py
```diff
@@ -514,11 +523,13 @@ def _create_resolver_script(self):
...
+        self._init_session_code = ''
```
```suggestion
```

frida_tools/tracer.py
```diff
@@ -573,6 +584,19 @@ def stop(self):
             self._script = None
 
     def _create_trace_script(self):
+        init_session_code = ""
+        for path in init_session_paths:
```
```suggestion
        for path in self._init_session_paths:
```

frida_tools/tracer.py
```diff
@@ -581,6 +605,10 @@ def _create_trace_script(self):
 var pending = [];
 var timer = null;
 
+state.newline = String.fromCharCode(10);
```
```suggestion
```

Can we escape this instead of requiring use of a constant?

frida_tools/tracer.py
```diff
@@ -678,7 +706,11 @@ def _create_trace_script(self):
         return {};
     }
 }
-"""
+
+function initializeSession() {
+    eval("%(init_code)s");
```
oleavr: This breaks if a session script contains double quotes. It would be good to pass it in through RPC instead. (Like we discussed on the other PR.)

frida_tools/tracer.py
mosherubin: It is crucial not to write any string containing a slash and an 'n'.  Calling String.fromCharCode(10) enables us to avoid such strings.  What did you have in mind?

mosherubin:
Hi oleavr,

Just to let you know I've been absent from this PR due to an important deadline at work, but hope to return to complete these PRs.  I have not forgotten :-)

Moshe

frida_tools/tracer.py
mosherubin: Both --init-session and --decorate help strings are now lowercase.

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: I need your help here.

I modified the code to execute all --init-session JS files in the `rpc.exports: init` function.  Unfortunately, when running the code, the `init` function is not being invoked.  I will commit the non-working code for your examination.

Here is the relevant `tracer.py` code:

```
    def _create_trace_script(self):
        init_session_code = ""
        for path in self._init_session_paths:
            with open(path, 'rt') as fo:
                lines = fo.readlines()

            init_session_code += "\n".join(lines)
            init_session_code += "\n"

        return """'use strict';

var started = Date.now();
var handlers = {};
var state = {};
var pending = [];
var timer = null;

state.newline = String.fromCharCode(10);
state.bar = "I am BAR!";    //debug

rpc.exports = {
    init: function (stage, parameters) {
        %(init_code)s
    },
    dispose: function () {
        flush();
    },
```

Note:

- The line `init_session_code += "\n"` is necessary to separate two different JS file by a newline.

- The `init` function seems straightforward to me (below I will show you what is generated at runtime)

- I have inserted the line `state.bar = "I am BAR!";  //debug`.  This will be used to show that `state.bar` is set correctly and shows up in the test handler, while the `state.foo` variable in the --init-session code is NOT set.

For testing I used the following --init-session JS file (entitled `init-session.js`):

```
state.foo = 'I am FOO!';
```

I used the `msvcrt.dll!_lock` function handler for testing:

```
  onEnter: function (log, args, state) {
    log('_lock() [msvcrt.dll]');
    log('state.foo = "' + state.foo + '"')
    log('state.bar = "' + state.bar + '"')
  },
```

And finally, here is the command line I used to test everything:

```
frida-trace -p 17860 -i "msvcrt.dll!_lock" -S "D:\Junk\frida-test\init-session.js"
```

(For the record, I am using the `vlc.exe` application for testing.).

Now we come to the runtime results.  At lines 551-553 I inserted a dump of the string returned by self._create_trace_script(), so we can see exactly what the JS trace looks like:

```
D:\Junk\frida-test>frida-trace -p 17860 -i "msvcrt.dll!_lock" -S "D:\Junk\frida-test\init-session.js"
Instrumenting functions...
----- _create_trace_script()  ------------------------------
'use strict';

var started = Date.now();
var handlers = {};
var state = {};
var pending = [];
var timer = null;

state.newline = String.fromCharCode(10);
state.bar = "I am BAR!";

rpc.exports = {
    init: function (stage, parameters) {
        state.foo = 'I am FOO!';
    },
    dispose: function () {
        flush();
    },
    add: function (targets) {
        targets.forEach(function (target) {
            var h = [parseHandler(target)];
            var name = target.name;
            var targetAddress = target.absolute_address;
            target = null;

            handlers[targetAddress] = h;

            function invokeCallback(callback, context, param) {
                if (callback === undefined)
                    return;

                var timestamp = Date.now() - started;
                var threadId = context.threadId;
                var depth = context.depth;

                function log(message) {
                    emit([timestamp, threadId, depth, targetAddress, message]);
                }

                callback.call(context, log, param, state);
            }

            try {
                Interceptor.attach(ptr(targetAddress), {
                    onEnter: function (args) {
                        invokeCallback(h[0].onEnter, this, args);
                    },
                    onLeave: function (retval) {
                        invokeCallback(h[0].onLeave, this, retval);
                    }
                });
            } catch (e) {
                send({
                    from: '/targets',
                    name: '+error',
                    payload: {
                        message: 'Skipping "' + name + '": ' + e.message
                    }
                });
            }
        });
    },
    update: function (targets) {
        targets.forEach(function (target) {
            handlers[target.absolute_address][0] = parseHandler(target);
        });
    }
};

function emit(event) {
    pending.push(event);

    if (timer === null)
        timer = setTimeout(flush, 50);
}

function flush() {
    if (timer !== null) {
        clearTimeout(timer);
        timer = null;
    }

    if (pending.length === 0)
        return;

    var items = pending;
    pending = [];

    send({
        from: '/events',
        name: '+add',
        payload: {
            items: items
        }
    });
}

function parseHandler(target) {
    try {
        return (1, eval)('(' + target.handler + ')');
    } catch (e) {
        send({
            from: '/targets',
            name: '+error',
            payload: {
                message: 'Invalid handler for "' + target.name + '": ' + e.message
            }
        });
        return {};
    }
}

------------------------------------------------------------
_lock: Loaded handler at "D:\\Junk\\frida-test\\__handlers__\\msvcrt.dll\\_lock.js"
Started tracing 1 function. Press Ctrl+C to stop.
           /* TID 0x42bc */
 12082 ms  _lock() [msvcrt.dll]
 12082 ms  state.foo = "undefined"
 12082 ms  state.bar = "I am BAR!"
 12083 ms  _lock() [msvcrt.dll]
 12083 ms  state.foo = "undefined"
 12083 ms  state.bar = "I am BAR!"
 12083 ms  _lock() [msvcrt.dll]
 12083 ms  state.foo = "undefined"
 12083 ms  state.bar = "I am BAR!"
 ...
 ```

Note:

- The `state.bar = "I am BAR!";` statement is correctly set in the `_lock` handler trace output.

- The `rpc.exports:init()` function does not seem to have been called, as the `state.foo` variable is undefined in the handler trace output.

Summary: Can you tell me why the `rpc.exports:init()` is not being invoked?

frida_tools/tracer.py
mosherubin: I need to examine this one carefully.  I'm committing everything else and will implement this one in parallel.

frida_tools/tracer.py
oleavr: I was thinking we could escape the user's `\n` instead of requiring them to use a constant.

frida_tools/tracer.py
oleavr: There isn't any magic that calls `init()`, we need to call it ourselves just like any other RPC method. (It's an application-level protocol that we implement in `frida-gadget` and `frida-inject`.)

oleavr:
Tweaked version landed as 8cb6ddae74ced0397264bdb45a832c716c2ab5d5. Thanks! :heart:

oleavr:
Released in 5.4.0 :tada:

PR #29, opened 2019-11-28
Populate <state> global variable with user data before beginning trace session

mosherubin:
The ability for the user to pass a session-specific string on the command line and have it stored in the <state> global variable before starting the trace will allow more dynamic control of handler behaviour.  This data-driven, rather than code-driven, model enables the user to modify what a handler will do without having to modify its code on a continual basis.

### DIFF
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -347,20 +347,16 @@ function excludeModule(pattern, workingSet) {
     return workingSet;
 }
 
-function parseExportsFunctionPattern (pattern) {
-    var m;
-    var f;
-    var res = pattern.split ('!');
-    
-    if (res.length == 1) {
+function parseExportsFunctionPattern(pattern) {
+    var res = pattern.split('!');
+    var m, f;
+    if (res.length === 1) {
         m = '*';
         f = res[0];
+    } else {
+        m = (res[0] === '') ? '*' : res[0];
+        f = (res[1] === '') ? '*' : res[1];
     }
-    else {
-        m = (res[0] == '' ? '*' : res[0]);
-        f = (res[1] == '' ? '*' : res[1]);
-    }
-    
     return {
         module: m,
         function: f
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -69,13 +69,7 @@ def main():
             self._decorate = options.decorate
             self._output = None
             self._output_path = options.output
-            self._userdata = self.escape_char(options.userdata, '"')
-
-        def escape_char(self, str, c):
-            if (str == None):
-                return ''
-            else:
-                return str.replace(c, '\\'+c)
+            self._userdata = options.userdata
 
         def _needs_target(self):
             return True
@@ -544,10 +538,7 @@ class Tracer(object):
         self._profile = profile
         self._script = None
         self._log_handler = log_handler
-        
-        self._userdata = userdata
-        if self._userdata == None:
-            self._userdata = ''
+        self._userdata = userdata if userdata is not None else ""
 
     def start_trace(self, session, runtime, ui):
         def on_create(*args):
@@ -609,9 +600,10 @@ var state = {};
 var pending = [];
 var timer = null;
 
-state.userdata = "%(userdata)s";
-
 rpc.exports = {
+    init: function (stage, parameters) {
+        state.userdata = %{userdata}s;
+    },
     dispose: function () {
         flush();
     },
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -54,6 +54,7 @@ def main():
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
             parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
+            parser.add_option("-u", "--userdata", help="Store user-defined data in state.userdata variable", metavar="STRING", type='string', default='')
             parser.add_option("-o", "--output", help="dump messages to file", metavar="OUTPUT", type='string')
             self._profile_builder = pb
 
@@ -68,6 +69,13 @@ def main():
             self._decorate = options.decorate
             self._output = None
             self._output_path = options.output
+            self._userdata = self.escape_char(options.userdata, '"')
+
+        def escape_char(self, str, c):
+            if (str == None):
+                return ''
+            else:
+                return str.replace(c, '\\'+c)
 
         def _needs_target(self):
             return True
@@ -75,7 +83,7 @@ def main():
         def _start(self):
             if self._output_path is not None:
                 self._output = OutputFile(self._output_path)
-            self._tracer = Tracer(self._reactor, FileRepository(self._reactor, self._decorate), self._profile, log_handler=self._log)
+            self._tracer = Tracer(self._reactor, FileRepository(self._reactor, self._decorate), self._profile, self._userdata, log_handler=self._log)
             try:
                 self._targets = self._tracer.start_trace(self._session, self._runtime, self)
             except Exception as e:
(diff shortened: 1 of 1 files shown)


frida_tools/tracer.py
```diff
@@ -514,12 +538,16 @@ def _create_resolver_script(self):
...
 class Tracer(object):
-    def __init__(self, reactor, repository, profile, log_handler=None):
+    def __init__(self, reactor, repository, profile, userdata, log_handler=None):
         self._reactor = reactor
         self._repository = repository
         self._profile = profile
         self._script = None
         self._log_handler = log_handler
+        
+        self._userdata = userdata
```
oleavr: Could be simplified to:

```suggestion
        self._userdata = userdata if userdata is not None else ""
```

frida_tools/tracer.py
```diff
@@ -581,6 +609,8 @@ def _create_trace_script(self):
 var pending = [];
 var timer = null;
 
+state.userdata = "%(userdata)s";
+
 rpc.exports = {
```
oleavr: Hmm, perhaps it would be cleaner to define an RPC export here named `init`, having the same signature as the described in the [Gadget docs](https://frida.re/docs/gadget/). This way we can call it right after loading the script, and there is no need to escape anything.

frida_tools/tracer.py
```diff
@@ -68,14 +69,21 @@ def _initialize(self, parser, options, args):
             self._decorate = options.decorate
             self._output = None
             self._output_path = options.output
+            self._userdata = self.escape_char(options.userdata, '"')
```
oleavr: How about parsing this as JSON instead?

frida_tools/tracer.py
```diff
@@ -54,6 +54,7 @@ def process_builder_arg(option, opt_str, value, parser, method, **kwargs):
                     type='string', action='callback', callback=process_builder_arg, callback_args=(pb.include_debug_symbol,))
             parser.add_option("-q", "--quiet", help="do not format output messages", action='store_true', default=False)
             parser.add_option("-d", "--decorate", help="Add module name to generated onEnter log statement", action='store_true', default=False)
+            parser.add_option("-u", "--userdata", help="Store user-defined data in state.userdata variable", metavar="STRING", type='string', default='')
```
```suggestion
            parser.add_option("-P", "--parameters", help="Store user-defined parameters in global parameters variable", metavar="JSON", type='string', default='')
```

For consistency with [frida-inject](https://github.com/frida/frida-core/blob/76172f64573ef80bc1293442750b41867e3157b3/inject/inject.vala#L22) and frida-gadget.

frida_tools/tracer.py
mosherubin: This was a challenging request, as I had to understand the concept of Python RPC exports etc., but I believe I got it in the end.  I have removed all escaping of `options.userdata`, inserting the following init() function in `rpc.exports`:

```
rpc.exports = {
    init: function (stage, parameters) {
        state.userdata = %{userdata}s;
    },
    dispose: function () {
        flush();
    },
```

Your idea is clever and clean, bypassing the need to be worried about the leading and trailing double quotes.  Please verify that it is what you meant.

frida_tools/tracer.py
mosherubin: Great minds think alike!  Here is newer code I have been using (modified per using `rpc.exports`:

```
        return """'use strict';

var started = Date.now();
var handlers = {};
var state = {};
var pending = [];
var timer = null;

state.newline = String.fromCharCode(10);

try {
    state.userdata_json_object = JSON.parse(state.userdata);
} catch (e) {
    // user data is not valid JSON, leave it as-is
}
```
We don't want to force the user to pass JSON data via the `-userdata` option, but we should make his life easier if he does.  So if a user knows he's passing JSON via `userdata`, he can expect to find a valid variable `state.userdata_json_object`, if the data is valid.  Of course, `state.userdata` exists in any case.

What are your thoughts?

frida_tools/tracer.py
mosherubin: Referring to your previous comment (i.e., parsing state.userdata as JSON), I assume this comment backtracks somewhat and recommends having a JSON and non-JSON user-provided data.

My original thought was that passing JSON data allows passing an unlimited amount of data to the system.  The following JSON would allow the user to pass data to two different handler sub-systems and unstructured data:

```
{
    "subsys-1": {
        "parm-1": 777,
        "parm-2": "a string"
    },
    "subsys-2": {
        "parm-1": 888,
        "parm-2": "another string"
    },
    "unstructured": "non-JSON unstructured text"
}
```

So JSON data is as flexible as we might want to be.  Having said that, I have no problem with two types of user data, JSON and non-JSON.

So should I give it a try?

frida_tools/tracer.py
mosherubin: Done.

frida_tools/tracer.py
mosherubin: Please see the next issue, I'm awaiting your decision.

frida_tools/tracer.py
oleavr: I'm leaning towards enforcing the same constraint that we enforce in frida-inject and frida-gadget: that the root has to be a JSON object. How about:

`-P '{"size": 42}'`
`-Z name:Joe`

Where `-Z` just adds a string property to the `parameters` object – which is an empty object unless you also specified `-P`.

So `-P '{"size": 42}' -Z name:Joe` would produce a `parameters` object looking like this:

```js
parameters = {
  name: 'Joe',
  size: 42
}
```

Is this on crack or does it make sense? Haven't spent a lot of time thinking about this.

frida_tools/tracer.py
```diff
@@ -582,6 +601,9 @@ def _create_trace_script(self):
 var timer = null;
 
 rpc.exports = {
+    init: function (stage, parameters) {
```
```suggestion
    init: function (stage, params) {
```

frida_tools/tracer.py
```diff
@@ -582,6 +601,9 @@ def _create_trace_script(self):
...
+        state.userdata = %{userdata}s;
```
```suggestion
        parameters = Object.freeze(params);
```

And adding `var parameters = {};` right before `var state = {};`. The thinking is that `parameters` should be treated as immutable input, and `state` is for mutable data only.

frida_tools/tracer.py
```diff
@@ -678,7 +700,7 @@ def _create_trace_script(self):
         return {};
     }
 }
-"""
+""" % {"userdata": self._userdata}
```
```suggestion
"""
```

So instead of passing it in here, we should call the `init()` method right after we have done `self._script.load()`:

```py
        self._script.exports.init(stage, parameters)
```

So we'll have to pass in `stage` and `parameters` to `start_trace()`:

```py
    def start_trace(self, session, runtime, stage, parameters, ui):
```

And the caller will need to compute `stage`, like we do in the REPL:

```py
        stage = 'early' if self._target[0] == 'file' and is_first_load else 'late'
```

(We should push this down into the `ConsoleApplication` base class.)

frida_tools/tracer.py
oleavr: Glad to hear. I've clarified a bit in the other comment, hope it makes it clearer. (Sorry for being so brief earlier.)

oleavr:
Implemented a slightly different solution in de8e34759badf7622521c20ff12e9720203473ad, which is consistent with our REPL. Thanks a lot for digging into this! :raised_hands:

mosherubin:
Oleavr,

A big thank you for making the last few steps to close PRs 27-29 on your own.  I salute and commend your collaborative style: you prefer that a PR submitter be guided and taught to make the correct changes on his/her own.  In my case, the back-and-forth on these last PR steps would have prevented getting the features out in the immediate future.

I have learned so much about Python and related topics from you, and I thank you for that.

It has been a pleasure working with you on Frida-related PRs.  One hears horror stories of other Open Source projects regarding the behavior of the primary maintainers to submitters.  If anyone is reading this, working with you on Frida and Frida-Tools has been a most wonderful experience for me -- I highly recommend collaborating with Oleavr on Frida (you will thank me for this advice!).  In fact, it has whetted my appetite for more, so this will not be the last you've heard from me! :-)

My next frida-tools-related tasks is to update the documentation with these last three PR new options (27-29), so stay tuned.

Wishing you a Happy New Year, and looking forward to more collaborations in the future,

Moshe

oleavr:
@mosherubin Thank you for your kind and generous words, and for having patience with my excessive pedantry/OCD! :blush: It was also a pleasure from my side, so am very much looking forward to future collaborations as well!

Happy New Year! :tada:

PR #30, opened 2019-12-09
handle log file UnicodeEncodeError

iddoeldor:


frida_tools/repl.py
```diff
@@ -127,7 +127,10 @@ def _initialize(self, parser, options, args):
         def _log(self, level, text):
             ConsoleApplication._log(self, level, text)
             if self._logfile is not None:
-                self._logfile.write(text + "\n")
+                try:
+                    self._logfile.write(text + "\n")
```
oleavr: Nice catch! Do we actually handle UTF-8 correctly here from before? A bit worried that it's defaulting to ASCII. (Haven't looked into that.)

frida_tools/repl.py
```diff
@@ -127,7 +127,10 @@ def _initialize(self, parser, options, args):
...
+                except UnicodeEncodeError:
+                    self._logfile.write(text.encode('ascii', 'ignore').decode('ascii') + '\n')
```
oleavr:
```suggestion
                    self._logfile.write(text.encode('ascii', 'ignore').decode('ascii') + "\n")
```

As per our [style guide](https://github.com/frida/frida/blob/master/CONTRIBUTING.md#rule-54).

frida_tools/repl.py
iddoeldor: Sorry for the delay, I didn't notice the comment.

I don't fully understand your comment, the 'ignore' part will tell it to just skip those characters. 

From the python docs:
```py
>>> u = unichr(40960) + u'abcd' + unichr(1972)
>>> u.encode('utf-8')
'\xea\x80\x80abcd\xde\xb4'
>>> u.encode('ascii')
Traceback (most recent call last):
  File "<stdin>", line 1, in ?
UnicodeEncodeError: 'ascii' codec can't encode character '\ua000' in position 0: ordinal not in range(128)
>>> u.encode('ascii', 'ignore')
'abcd'
>>> u.encode('ascii', 'replace')
'?abcd?'
>>> u.encode('ascii', 'xmlcharrefreplace')
'&#40960;abcd&#1972;'
```

frida_tools/repl.py
iddoeldor: Sorry.

oleavr:
Thanks for looking into this! :heart: This is unfortunately addressing the symptom, and not the cause. Just fixed in 1023dd56fbb5206e47b2ab35d2764a596203af38.

oleavr:
Fix released in 5.4.0 :tada:

PR #31, opened 2020-02-28
BUGFIX: Handle "--options-file=<path>" in options file

mosherubin:
The previous release of Frida introduced the "--options-file=<path>"
option.  This option allows placing any command line options and flags
in a text file and specifying the text file's path on the command line.

The short form of "--options-file" is "-O".

The code is meant to support "--options-file=" options embedded within
an options file, like so:

Command line:

    frida-trace -p 5620 --options-file=options-file.1.txt

options-file.1.txt:

    --options-file=options-file.2.txt

options-file.2.txt:

    -i "str*"
    -X "KERNELBASE.dll"

This should produce results equivalent to the command line:

    frida-trace -p 5620 -i "str*" -X "KERNELBASE.dll"

There is a bug whereby the "--options-file=<path>" on the command line
is processed correctly, but the embedded "--options-file=<path>" is not.

Note that there is no problem if the user uses the short form "-O" in
the embedded options file, only if the long form "--options-file=<path>"
is used.

The fix is to call the "normalize_options_file_args()" function both
when processing the command line *and* when processing the options
files themselves.

In addition, several of the functions were renamed for clarity reasons.

### DIFF
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -453,11 +453,11 @@ class ConsoleApplication(object):
 
 
 def compute_real_args(parser):
-    real_args = normalize_raw_args(sys.argv[1:])
+    real_args = normalize_options_file_args(sys.argv[1:])
 
     files_processed = set()
     while True:
-        offset = find_arg_file_offset(real_args, parser)
+        offset = find_options_file_offset(real_args, parser)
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks a lot! 🙌

PR #35, opened 2020-05-14
Take config dir from $XDG_CONFIG_HOME if it exists

gergesh:


frida_tools/repl.py
```diff
@@ -666,7 +666,10 @@ def _load_codeshare_script(self, uri):
                     return script
 
         def _get_or_create_config_dir(self):
-            config_dir = os.path.join(os.path.expanduser('~'), '.frida')
+            if os.getenv('XDG_CONFIG_HOME') is not None:
+                config_dir = os.path.join(os.getenv('XDG_CONFIG_HOME'), 'frida')
+            else:
+                config_dir = os.path.join(os.path.expanduser('~'), '.frida')
```
oleavr, line 672:
```suggestion
            xdg_home = os.getenv("XDG_CONFIG_HOME")
            if xdg_home is not None:
                config_dir = os.path.join(xdg_home, "frida")
            else:
                config_dir = os.path.join(os.path.expanduser("~"), ".frida")
```

PR #46, opened 2021-01-06
Add instructions for how to load custom frida-java-bridge code

Happyholic1203:
Many thanks to @oleavr for giving me the idea how to do this, and this PR might not be the best way to do it but I got it working.

Please advise if there's any better way to do this!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,19 @@ For example, on Windows, assuming you clone to `C:\src`:
 
 Now when you run frida.exe, these scripts and any changes you make will be
 referenced instead!
+
+### Loading your custom frida-java-bridge
+
+Add the following lines to `agents/tracer/agent.ts`:
+
+    export {};  // to get around TS2451: Cannot redeclare block-scoped variable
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,3 +18,19 @@ For example, on Windows, assuming you clone to `C:\src`:
 
 Now when you run frida.exe, these scripts and any changes you make will be
 referenced instead!
+
+### Loading your custom frida-java-bridge
+
+Add the following lines to `agents/tracer/agent.ts`:
+
+    export {};  // to get around TS2451: Cannot redeclare block-scoped variable
+    const Java = require('frida-java-bridge');
```
oleavr, line 27: I think this should work and be a lot simpler:

```suggestion
    Object.defineProperty(global, 'Java', { value: require('frida-java-bridge') });
```

oleavr (changes requested):
So so sorry for the delay. I'm terrible at context-switching!

PR #47, opened 2021-02-17
Fix incorrect cached resolver name

dki:
When an ObjC resolver is created, it is incorrectly cached under  `this.cachedModuleResolver` instead of `this.cachedObjcResolver`.  This results in cryptic errors if a trace combines both ObjC and function tracing when the ObjC resolver is initialized first, since the wrong resolver ends up being used to resolve the functions.


oleavr:
Whoa, nice catch! Thanks! 🙌

PR #50, opened 2021-05-07
Enable editor editing in prompt toolkit

yotamN:
My one liners were getting too long and apparently prompt toolkit have a feature to edit the prompt in the editor just like in a normal shell. I just pass the argument so it would enable it


oleavr:
Ooh nice, thank you!

PR #52, opened 2021-05-08
Abort return values completion instead of crashing

yotamN:
The current completion implementation is simple and can't know the
return type of the function without calling it (a big no no). Until
there is a better solution we should abort the completion instead of
showing an annoying error message to the user.


oleavr:
Thanks! 🙏

PR #54, opened 2021-05-11
Add fish completion

yotamN:
This is really just a POC for the command completion, ideally most of the flags will be automatically completed from the man pages once we have them and only the parameters with values (for example `--attach-pid`) will have a special configuration.

Additionally, I'm not sure the best way of installing the completion through pip. The path differs depending on if the user installed it as root or not and some users might not even want it.

- [ ] Write man pages
- [ ] Figure out how to install completions (also relevant to man pages)
- [x] Add missing commands

### DIFF
--- a/completions/meson.build
+++ b/completions/meson.build
@@ -1,0 +1,3 @@
+datadir = get_option('datadir')
+completions_dir = join_paths(datadir, 'fish', 'vendor_completions.d')
+install_data(['frida.fish'], install_dir: completions_dir)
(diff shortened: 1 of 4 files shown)


yotamN:
There is still a small problem with `frida-join` and of course I haven't gotten to write the manuals yet but I think it is good enough to be useful so I would love for it to get merge. Maybe after fixing the last issue I will try to merge it into Fish itself

yotamN:
I've added completions installation to the Meson script but I'm still not sure how to build with Meson

completions/meson.build
```diff
@@ -0,0 +1,3 @@
+datadir = get_option('datadir')
+completions_dir = join_paths(datadir, 'fish', 'vendor_completions.d')
```
oleavr, line 2:
```suggestion
completions_dir = get_option('datadir') / 'fish' / 'vendor_completions.d'
```

oleavr (approved):
Yay, this is great! 🤘

PR #60, opened 2021-08-29
Use UTF-8 encoding when writing tracer handler

AkiraNoSushi:
Python tries to write the handler using cp1252 on Windows, which triggers an UnicodeEncodeError exception when it encounters chinese characters. Specifying UTF-8 as our encoding fixes the issue.

### DIFF
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -636,7 +636,7 @@ class FileRepository(Repository):
             handler_dir = os.path.dirname(handler_file)
             if not os.path.isdir(handler_dir):
                 os.makedirs(handler_dir)
-            with open(handler_file, 'w') as f:
+            with open(handler_file, 'w', encoding='utf-8') as f:
                 f.write(handler)
             self._notify_create(target, handler, handler_file)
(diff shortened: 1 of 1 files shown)


frida_tools/tracer.py
```diff
@@ -636,7 +636,7 @@ def ensure_handler(self, target):
             handler_dir = os.path.dirname(handler_file)
             if not os.path.isdir(handler_dir):
                 os.makedirs(handler_dir)
-            with open(handler_file, 'w') as f:
+            with open(handler_file, 'w', encoding='utf-8') as f:
```
oleavr, line 639: Should use `codecs.open()` here as we still support Python 2.7. (Will be dropped soon though but in the meantime we don't want to break it.)

oleavr (changes requested):
Thanks!

frida_tools/tracer.py
oleavr, line 639:
```suggestion
            with codecs.open(handler_file, 'w', 'utf-8') as f:
```

PR #68, opened 2021-10-05
Add frida-apk

meme:
This is a new tool used for marking Android applications as debuggable so they may be used by Frida for e.g. injecting the gadget.

### DIFF
--- a/frida_tools/apk.py
+++ b/frida_tools/apk.py
@@ -1,0 +1,311 @@
+# -*- coding: utf-8 -*-
+from enum import IntEnum
+from io import BufferedReader
+from zipfile import ZipFile
+import os
+import struct
+
+
+def main() -> None:
+    import frida
+    from frida_tools.application import ConsoleApplication, infer_target, expand_target
+
+    class ApkApplication(ConsoleApplication):
+        def _usage(self):
+            return "usage: %prog [options] path.apk"
+
+        def _add_options(self, parser):
+            parser.add_option("-o", "--output", help="output path", metavar="OUTPUT", type="string")
+
+        def _needs_device(self):
+            return False
+
+        def _initialize(self, parser, options, args):
+            self._output_path = options.output
+
+            if len(args) < 1:
+                parser.error("path must be specified")
+            self._path = args[0]
+            if not self._path.endswith(".apk"):
+                parser.error("path must end in .apk")
+
+            if self._output_path is None:
+                self._output_path = self._path.replace(".apk", ".d.apk")
+
+        def _start(self):
+            try:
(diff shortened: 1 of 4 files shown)


frida_tools/apk.py
```diff
@@ -0,0 +1,313 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
```
oleavr, line 2: This should be omitted, since this file is Python 3.x only.

frida_tools/apk.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+from io import BufferedReader
+import struct
+from enum import IntEnum
+import os
+from zipfile import ZipFile
```
oleavr, line 9: These should be sorted alphabetically, by module name.

frida_tools/apk.py
```diff
@@ -0,0 +1,313 @@
...
+        # namespace for the debuggable tag we could scan the strings section
+        # for the AndroidManifest schema tag
+        if self.namespace == -1:
+            raise BadHeader()
+
+        chunk_data = bytearray(self.header.chunk_data)
+
+        resource_size = 8
+        resource_type = ResourceType.BOOL
+        resource_data = -1  # True
```
oleavr, line 79: What does this `# True` comment try to explain? (Or is it an artifact left from an earlier iteration?)

frida_tools/apk.py
```diff
@@ -0,0 +1,313 @@
...
+                    pass
+                else:
+                    oz.writestr(info.filename, f.read(), info.compress_type)
+
+
+def main() -> None:
+    import frida
+    from frida_tools.application import ConsoleApplication, infer_target, expand_target
+
+    class KillApplication(ConsoleApplication):
```
oleavr, line 275: Need to substitute the name here.

frida_tools/apk.py
```diff
@@ -0,0 +1,313 @@
...
+                        data.extend(header.chunk_data)
+
+                    oz.writestr(info.filename, bytes(data), info.compress_type)
+                elif info.filename.startswith("META-INF"):
+                    pass
+                else:
+                    oz.writestr(info.filename, f.read(), info.compress_type)
+
+
+def main() -> None:
```
oleavr, line 271: Would be good to move this function to the top — should be able to start reading from the top, where classes/components used by the code at the top follows below it, from highest to lowest level details. (Often referred to as [newspaper code structure](https://kentcdodds.com/blog/newspaper-code-structure)).

oleavr (changes requested):
Wow, nice! 🔥

PR #73, opened 2022-01-24
Use correct directory for each file type

yotamN:
The current implementation did not fit well in any operation system. There was some work on support xdg directories but in reality it was rarely used because most people use the default XDG directory without setting the environment variable.

This patch support XDG directories properly in addition of supporting the native folder location in Windows (someone with a bit more knowledge about MacOS could improve it there too). Not only that but I made sure to separate the directories to config,  state and data directories just like in the XDG specification (there are more directories that were ignored for now).

NOTICE: It might seems like a lot of code but I had to make sure that I'm migrating the old files properly and I kept the `_get_or_create_config_dir` even though we don't need it anymore so that same mistake won't happen in the future. I might also want to use it soon in a different patch

### DIFF
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -605,6 +605,47 @@ class ConsoleApplication(object):
 
         return result[0]
 
+    def _get_default_frida_dir(self):
+        return os.path.join(os.path.expanduser('~'), '.frida')
+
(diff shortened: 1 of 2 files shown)


oleavr (approved):
Nice work. Thanks!

PR #75, opened 2022-01-26
Migrate from deprecated optparse to argparse

yotamN:
Optparse is deprecated since Python 3.2 (more than 10 years ago!) so it's time to migrate to the newer argparse module that is still supported.

There wasn't much to change because argparse tries as best as it can to be backward compatible. Mostly what was changed is callback arguments and handling of extra arguments.

I also deleted redundant arguments that were the default anyway.

There's still more testing to do, that is why I marked it as WIP but I wanted to get feedback as soon as possible and I hope there won't be any more changes, except bug fixes of course.

### DIFF
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals, print_function
 
+import argparse
 import codecs
 import collections
 import errno
 import numbers
-from optparse import OptionParser
 import os
 import platform
 import re
@@ -70,6 +70,23 @@ def await_enter(reactor):
         print("")
 
 
+def deserialize_relay(value):
+    address, username, password, kind = value.split(",")
+    return frida.Relay(address, username, password, kind)
+
+
+def create_target_parser(type):
+    def parse_target(value):
+        if type == 'file':
+            value = [value]
+        elif type == 'gated':
+            value = re.compile(value)
+        elif type == 'pid':
+            value = int(value)
+        return (type, value)
+    return parse_target
+
+
 class ConsoleState:
     EMPTY = 1
     STATUS = 2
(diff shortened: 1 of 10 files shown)


yotamN:
Tested a bit more, of course would be glad for more testing but I hope it's okay

frida_tools/application.py
```diff
@@ -5,7 +5,7 @@
 import collections
 import errno
 import numbers
-from optparse import OptionParser
+import argparse
```
oleavr, line 8: These should be kept alphabetically sorted.

frida_tools/application.py
```diff
@@ -70,6 +70,22 @@ def await_enter(reactor):
...
+    return frida.Relay(address, username, password, kind)
+
+
+def create_target_parser(type):
+    def parse_target(value):
+        if type == "file":
+            value = [value]
+        elif type == "gated":
+            value = re.compile(value)
+        elif type == "pid":
```
oleavr, line 84: We currently use single-quoted strings for enum/symbol values, and double-quoted for everything else. So would be good to use that here too.

frida_tools/application.py
```diff
@@ -86,83 +102,52 @@ def __init__(self, run_until_return=await_enter, on_stop=None):
...
+        parser.add_argument('--version', action='version', version=frida.__version__)
 
         self._add_options(parser)
 
         real_args = compute_real_args(parser)
-        (options, args) = parser.parse_args(real_args)
+        options = parser.parse_args(real_args)
+
+        # handle scripts that don't need a target
+        if not(hasattr(options, 'args')):
```
oleavr, line 145:
```suggestion
        if not hasattr(options, 'args'):
```

oleavr (changes requested):
Ahh sweet. Thanks for doing this!

frida_tools/application.py
yotamN: Fixed, maybe we could add isort to the CI? Maybe with the migration to Python 3 we can add Black

frida_tools/application.py
yotamN: Fixed

frida_tools/application.py
yotamN: Fixed but a weird convention isn't it? Maybe it would be better to create the appropriate enums in the Python bindings?

frida_tools/application.py
oleavr: That would be great!

frida_tools/application.py
oleavr: Yeah I agree. Would be good to switch all quotes to one style. (Whichever you prefer, I'm fine with either.)

PR #76, opened 2022-01-26
Add a toggle to wrap every input in Java.perform

### LINKED ISSUE #74: Auto insert Java.perform when connecting to an Android application
I understand that you might not always want to attach a thread to the JVM with Frida so `Java.perform` is being used to make it explicit but in the repl it gets a bit tiresome to type it out each time I want to run something. Assuming there isn't a risk with doing automatically I suggest adding a feature that will wrap the repl prompt with `Java.perform`.

Of course it will be configurable with a command from the repl (`%autoperform` and `%noautoperform` I think) and can be off by default if it seems risky (in that case we should probably add a prompt to tell the user about this feature). What do you think @oleavr?

yotamN:
The patch make it off by default of course because there are still a few things to iron out. Mostly I need to be sure that there is no risk of doing it and to figure out how to make the variables globals.

Fixes #74

### DIFF
--- a/frida_tools/repl.py
+++ b/frida_tools/repl.py
@@ -494,7 +494,7 @@ def main():
             elif state:
                 self._print("autoperform is only available in Java processes")
 
-        def _is_java_avilable(self):
+        def _is_java_available(self):
             script = self._session.create_script(name="java_check", source="rpc.exports.javaAvailable = () => Java.available;", runtime=self._runtime)
             script.load()
             return script.exports.java_available()
--- a/frida_tools/repl.py
+++ b/frida_tools/repl.py
@@ -495,7 +495,7 @@ def main():
                 self._print("autoperform is only available in Java processes")
 
         def _is_java_avilable(self):
-            script = self._session.create_script(name="java_check", source="rpc.exports.javaAvailable = function() {return Java.available}", runtime=self._runtime)
+            script = self._session.create_script(name="java_check", source="rpc.exports.javaAvailable = () => Java.available;", runtime=self._runtime)
             script.load()
             return script.exports.java_available()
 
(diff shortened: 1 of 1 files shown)


frida_tools/repl.py
```diff
@@ -335,6 +340,8 @@ def _process_input(self, reactor):
                         if expression.startswith("%"):
                             self._do_magic(expression[1:].rstrip())
                         else:
+                            if self._autoperform:
+                                expression = f"Java.performNow(() => {{return {expression}\n/**/}});"
```
oleavr, line 344:
```suggestion
                                expression = f"Java.performNow(() => {expression});"
```

Could this be simplified to this perhaps? (Untested)

yotamN: Since I'm not actually parsing the code but just concatenating strings I tried to use the most versatile expression I could think of.

* The double curly braces is escaping of a single curly brace because this is inside a Python 3 string format. But now that I think about it, this is a problem because we still support Python 2, I will fix it.
* The line break is used so we won't fail if the inserted code ends with a line comment
* The `/**/` section is used so we won't fail if the inserted code ends with an open multi line comment

frida_tools/repl.py
```diff
@@ -471,6 +481,27 @@ def _reload(self):
...
+
+        def _set_autoperform(self, state):
+            if self._is_java_avilable():
+                self._autoperform = state
+                self._refresh_prompt()
+            elif state:
+                self._print("autoperform is only available in Java processes")
+
+        def _is_java_avilable(self):
+            script = self._session.create_script(name="java_check", source="rpc.exports.javaAvailable = function() {return Java.available}", runtime=self._runtime)
```
oleavr:
```suggestion
            script = self._session.create_script(name="java_check", source="rpc.exports.javaAvailable = () => Java.available;", runtime=self._runtime)
```

PR #77, opened 2022-01-27
Add help command that displays list of commands

yotamN:
Together with some refactor to make it possible, I've added an help message (and an unload message that was declared but not implemented).

I've developed it over my other patch of auto java perform because it touched the same code area and I wanted to minimize merge conflicts later, hope it's okay.

### DIFF
--- a/frida_tools/_repl_magic.py
+++ b/frida_tools/_repl_magic.py
@@ -1,0 +1,112 @@
+import abc
+import json
+
+
+class Magic(abc.ABC):
+    @property
+    def description(self):
+        return "no description"
+
+    @abc.abstractproperty
+    def required_args_count(self):
+        pass
+
+    def execute(self, repl, args):
+        pass
+
+
+class Resume(Magic):
+    @property
+    def description(self):
+        return "resume execution of the spawned process"
+
+    @property
+    def required_args_count(self):
+        return 0
+
+    def execute(self, repl, args):
+        repl._reactor.schedule(lambda: self._resume())
+
+
+class Reload(Magic):
+    @property
+    def description(self):
+        return "reload (i.e. rerun) the script that was given as an argument to the REPL"
+
+    @property
+    def required_args_count(self):
+        return 0
+
+    def execute(self, repl, args):
+        try:
+            repl._perform_on_reactor_thread(lambda: self._load_script())
+            return True
+        except Exception as e:
+            repl._print("Failed to load script: {}".format(e))
+            return False
+
+
+class Unload(Magic):
+    @property
+    def required_args_count(self):
+        return 0
+
+    def execute(self, repl, args):
+        repl._unload_script()
+
+
+class Autoperform(Magic):
+    @property
+    def description(self):
+        return "receive on/off as first and only argument, when switched on will wrap any REPL code with Java.performNow()"
+
+    @property
+    def required_args_count(self):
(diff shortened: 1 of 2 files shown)


frida_tools/_magics.py
```diff
@@ -0,0 +1,111 @@
+import abc
+import json
+
```
oleavr, line 3: Shouldn't there be two blank lines here according to PEP-8? (I might me misremembering but at least this is the convention I've used — so hope I didn't get it wrong 😅)

frida_tools/_magics.py
```diff
@@ -0,0 +1,111 @@
...
+        return 0
+
+    def execute(self, repl, args):
+        repl._reactor.schedule(lambda: self._resume())
+
+
+class Reload(Magic):
+    @property
+    def description(self):
+        return "reload (i.e. rerun) the script that was given as an argument to the repl"
```
oleavr, line 33:
```suggestion
        return "reload (i.e. rerun) the script that was given as an argument to the REPL"
```

frida_tools/_magics.py
```diff
@@ -0,0 +1,111 @@
...
+        pass
+
+    def execute(self, repl, args):
+        pass
+
+
+class Resume(Magic):
+    @property
+    def description(self):
+        return "sesume execution of the attached process"
```
oleavr, line 20:
```suggestion
        return "resume execution of the spawned process"
```

frida_tools/_magics.py
```diff
@@ -0,0 +1,111 @@
...
+        return 0
+
+    def execute(self, repl, args):
+        repl._unload_script()
+
+
+class Autoperform(Magic):
+    @property
+    def description(self):
+        return "receive on/off as first and only argument, when switched on will wrap any repl code with Java.PerformNow"
```
oleavr, line 60:
```suggestion
        return "receive on/off as first and only argument, when switched on will wrap any REPL code with Java.performNow()"
```

frida_tools/_magics.py
```diff
@@ -0,0 +1,111 @@
...
+                const _endTime = Date.now();
+                console.log('Time: ' + (_endTime - _startTime) + ' ms.');
+                return _result;
+            }})();'''.format(expression=json.dumps(" ".join(args))))
+
+
+class Help(Magic):
+    @property
+    def description(self):
+        return "print a list of available repl commands"
```
oleavr, line 93:
```suggestion
        return "print a list of available REPL commands"
```

frida_tools/_magics.py
```diff
@@ -0,0 +1,111 @@
+import abc
```
oleavr, line 1: Perhaps we should name this file `_repl_magic` or something? (Plural sounds a bit strange so perhaps singular is better?)

oleavr (changes requested):
Wow, nice. Thanks!

yotamN:
Fixed everything and rebased

oleavr:
Thanks! Conflicts ⚔️

yotamN:
Fixed

oleavr:
Conflicts ⚔️

yotamN:
rebased again

starplanet:
There is a bug that _repl_magic.py is not listed in meson.build file. It will cause frida to fail to compile

PR #78, opened 2022-01-27
Enable loading multiple scripts in the repl

yotamN:
Instead of only loading one file, let the user pass the --load flag multiple times and concatenate them together.

### DIFF
--- a/frida_tools/repl.py
+++ b/frida_tools/repl.py
@@ -67,7 +67,7 @@ def main():
                 self._rpc_complete_server = start_completion_thread(self)
 
         def _add_options(self, parser):
-            parser.add_argument("-l", "--load", help="load SCRIPT", metavar="SCRIPT", dest="user_script")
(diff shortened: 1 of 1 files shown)


frida_tools/repl.py
```diff
@@ -197,13 +195,16 @@ def _stop(self):
             if not self._quiet:
                 self._print("\nThank you for using Frida!")
 
+        def _get_script_name(self, path):
+            return os.path.splitext(os.path.basename(path))[0]
```
oleavr, line 199: This should go below the code that is using it. (Newspaper code structure.)

oleavr (changes requested):
Yay, great feature. Thanks for doing this!

frida_tools/repl.py
yotamN: Fixed and rebased against main

oleavr:
Thanks! Afraid we have some conflicts now ⚔️

yotamN:
I rebased

PR #79, opened 2022-01-27
Print the connected device in the repl

yotamN:
It can be a bit confusing when there are multiple devices to know which one you are actually working on. I've thought about it when I saw a question relating to this in the Telegram group.

### DIFF
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -343,6 +343,7 @@ class ConsoleApplication(object):
                 return
         else:
             self._device = frida.get_local_device()
+        self._on_device_found()
         self._device.on('output', self._schedule_on_output)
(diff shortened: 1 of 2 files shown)


frida_tools/application.py
```diff
@@ -605,6 +606,9 @@ def work():
 
         return result[0]
 
+    def _on_device_found(self):
+        pass
```
oleavr, line 610: Would be good to move this to just before `_on_device_lost`.

oleavr (changes requested):
Nice, thanks!

frida_tools/application.py
yotamN: Fixed and rebased against main

PR #81, opened 2022-01-30
TS4 compatibility: annotate catch errors as 'any' type

jevinskie:
I noticed this was needed when trying to build the tracer agent without the checked in package-lock.json


oleavr:
Thanks!

PR #82, opened 2022-02-02
Add exec magic command to the REPL

yotamN:
Sometimes you want to execute a multiline script after you've already opened the REPL and you don't want to close it and lose all state. The exec command will let the user execute a file from the REPL while running, just like typing the code manually.

Again I'm basing this patch on the magics refactors so it won't cause too much conflicts

### DIFF
--- a/frida_tools/_repl_magic.py
+++ b/frida_tools/_repl_magic.py
@@ -1,5 +1,7 @@
 import abc
+import codecs
 import json
+import os
 
 
 class Magic(abc.ABC):
@@ -66,6 +68,29 @@ class Autoperform(Magic):
 
     def execute(self, repl, args):
         repl._autoperform_command(args[0])
+
+
+class Exec(Magic):
+    @property
+    def description(self):
+        return "execute the given file path in the context of the currently loaded scripts"
+
+    @property
+    def required_args_count(self):
+        return 1
+
+    def execute(self, repl, args):
+        if not os.path.exists(args[0]):
+            repl._print("Can't read the given file because it does not exist")
+            return
+
+        try:
+            with codecs.open(args[0], 'rb', 'utf-8') as f:
+                if not repl._eval_and_print(f.read()):
+                    repl._errors += 1
+        except PermissionError:
+            repl._print("Can't read the given file because of a permission error")
+
 
 
 class Time(Magic):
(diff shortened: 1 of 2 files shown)


frida_tools/repl.py
```diff
@@ -78,6 +81,7 @@ def _add_options(self, parser):
             parser.add_argument("-o", "--output", help="output to log file", dest="logfile")
             parser.add_argument("--eternalize", help="eternalize the script before exit", action='store_true', dest="eternalize", default=False)
             parser.add_argument("--exit-on-error", help="exit with code 1 after encountering any exception in the SCRIPT", action='store_true', dest="exit_on_error", default=False)
+            parser.add_argument("--auto-perform", help="wrap entered code with Java.perform", action='store_true', dest="autoperform", default=False)
```
oleavr, line 84: This is very specific to one runtime, while the name doesn't suggest it. I'm wondering if we should make this more generic, so it also does the right thing when in e.g. an iOS app, where `ObjC.schedule(ObjC.mainQueue, ...)` could be used. I'm a bit reluctant to add an option for this though, and tempted to just do it automatically, with a way to opt out. (Like a quick way to switch the current "context", and have it indicated by the prompt perhaps?)

yotamN: Notice this is a code from a different PR #76, I just based this PR on the other one to reduce conflicts.

Regarding the comment, I'm afraid that because the auto perform wrapping isn't perfect it should not be enabled by default, at least not with the current implementation. Variables declared with `var foo` or `let bar` for examples won't survive between different `Java.perform` calls. But we can make it wrap ObjC as well but I don't have an iOS/MacOS device available so won't be able to do it myself :(

oleavr:
Conflicts ⚔️

yotamN:
Rebased

frida_tools/_repl_magic.py
```diff
@@ -68,6 +70,29 @@ def execute(self, repl, args):
...
+
+        try:
+            with codecs.open(args[0], 'rb', 'utf-8') as f:
+                if not repl._eval_and_print(f.read()):
+                    repl._errors += 1
+        except PermissionError:
+            repl._print("Can't read the given file because of a permission error")
+
+
+
```
oleavr, line 95:
```suggestion
```

oleavr:
Thanks!

PR #88, opened 2022-02-11
Fix relays argument

yotamN:
The relays were put correctly in the parsing result object but was not used correctly in the class.


oleavr (approved):
Thanks!

PR #89, opened 2022-02-11
Add unit tests to device arguments parsing

yotamN:
There is still a lot to cover but those patches should make it easier with the rest


oleavr:
Sorry, I probably shouldn't have merged the other PR before this one. Conflicts :crossed_swords:

yotamN:
Rebased

yotamN:
Added another test for multiple device flags

PR #91, opened 2022-02-17
Add autoreload flag that is on by default

### LINKED ISSUE #83: Add a way to disable auto reload
Auto reload is great but the way it's implemented it deletes any state in the REPL. I propose to add a flag to disable it.
I personally think auto reload should be off by default but because we can't change it without breaking backwards compatibility I suggest adding the flag `--no-auto-reload`. If we want to eventually make it the default I also suggest adding `--auto-reload` flag that does nothing so the migration will be easier in the future.

We should also have a way to change the settings from the REPL which mean a `%autoreload {on/off}` command.

I plan on implementing it after my magics refactor is merged.

yotamN:
It would have made more sense to make auto reload off by default but because we don't want to break backwards compatibility we keep it on by default. In order to ease any future migration the auto reload flag exist and will currently do nothing (except overriding any previous `--no-auto-reload` flag).

Resolves #83

### DIFF
--- a/frida_tools/_repl_magic.py
+++ b/frida_tools/_repl_magic.py
@@ -70,6 +70,33 @@ class Autoperform(Magic):
         repl._autoperform_command(args[0])
 
 
+class Autoreload(Magic):
+    _VALID_ARGUMENTS = ("on", "off")
+
+    @property
+    def description(self):
+        return "disable or enable auto reloading of script files"
+
+    @property
+    def required_args_count(self):
+        return 1
+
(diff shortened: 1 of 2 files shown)


oleavr:
@yotamN Thanks! Could you tweak the commit messages so they include a tag where appropriate? E.g. `repl: Fix xyz`. Otherwise this looks ready to merge.

yotamN:
Sure, fixed it

oleavr:
Thanks!

PR #94, opened 2022-03-02
Quiet mode improvements

otadmor:
These are not many changes. Each commit tell what it does.

### DIFF
--- a/frida_tools/repl.py
+++ b/frida_tools/repl.py
@@ -7,6 +7,7 @@ import signal
 import string
 import threading
 import sys
+import time
 
 
 def main():
@@ -47,6 +48,7 @@ def main():
             self._autoperform = False
             self._autoperform_option = False
             self._autoreload = True
+            self._quiet_start = None
 
             super(REPLApplication, self).__init__(self._process_input, self._on_stop)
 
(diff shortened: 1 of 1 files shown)


frida_tools/repl.py
```diff
@@ -82,6 +84,7 @@ def _add_options(self, parser):
             parser.add_argument("--eternalize", help="eternalize the script before exit", action='store_true', dest="eternalize", default=False)
             parser.add_argument("--exit-on-error", help="exit with code 1 after encountering any exception in the SCRIPT", action='store_true', dest="exit_on_error", default=False)
             parser.add_argument("--auto-perform", help="wrap entered code with Java.perform", action='store_true', dest="autoperform", default=False)
+            parser.add_argument("-t", "--timeout", help="seconds to wait before terminating on quiet mode", dest="timeout", default=0)
```
oleavr, line 87:
```suggestion
            parser.add_argument("-t", "--timeout", help="seconds to wait before terminating in quiet mode", dest="timeout", default=0)
```

oleavr (changes requested):
Thanks! Could you add a tag to the subject of your commit messages? E.g.:
`repl: Fix X`
`repl: Fix Y`
etc.
(This is a fairly new convention, but can be omitted if changes apply broadly.)

oleavr:
Actually, I'll take a quick stab at fixing this so we can get this released.

PR #95, opened 2022-03-05
fix `frida-apk` argument parsing

mhils:
This PR fixes the argument parsing for `frida-apk`, which previously failed as follows after the argparse migration:

```console
$ frida-apk
usage: frida-apk [options] path.apk
frida-apk: error: path must be specified
$ frida-apk path.apk
usage: frida-apk [options] path.apk
frida-apk: error: unrecognized arguments: path.apk
```

PR #97, opened 2022-03-06
`frida-apk`: Add support for ResourceMap, ordered inserts

mhils:
This PR makes `frida-apk` work with my Android 12 Pixel 3 device (stock ROM, no root).

See the 9285b273de68e3c95a22c713d9840bf9d21e6e8b commit message for a detailed explanation of the changes. I've taken the liberty to build this on top of #95, which hopefully should be uncontroversial.

FWIW I've built a small tool around this for certificate pinning (https://github.com/mitmproxy/android-unpinner). If you folks feel that the JDWP stuff is useful, please feel free to shamelessly copy it.

### DIFF
--- a/frida_tools/apk.py
+++ b/frida_tools/apk.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
+
+import os
+import struct
 from enum import IntEnum
 from io import BufferedReader
 from zipfile import ZipFile
-import os
-import struct
 
 
 def main() -> None:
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks, this is awesome! :heart_eyes:

> FWIW I've built a small tool around this for certificate pinning (https://github.com/mitmproxy/android-unpinner). If you folks feel that the JDWP stuff is useful, please feel free to shamelessly copy it.

Very cool! Re JDWP, Frida already has native support for gadget injection through JDWP. We haven't done a great job advertising it though, so it's a rather obscure feature even though it's already been around for a year. Basically if you use an Android `Device` to try to `spawn()` or `attach()` and Frida is unable to connect to a remote `frida-server`, it will attempt its non-rooted/jailed code-path. The gist of it happens in [injector.vala](https://github.com/frida/frida-core/blob/986e6a6a5270dd0c39c30472871cbe215c136352/src/droidy/injector.vala#L78-L245), which uses [jdwp.vala](https://github.com/frida/frida-core/blob/main/src/droidy/jdwp.vala) and [droidy-client.vala](https://github.com/frida/frida-core/blob/main/src/droidy/droidy-client.vala) for the heavy lifting. That said it currently assumes arm64, and given how few folks know about it it likely needs some love :blush:

mhils:
> Very cool! Re JDWP, Frida already has native support for gadget injection through JDWP. We haven't done a great job advertising it though, so it's a rather obscure feature even though it's already been around for a year.

The more you know... 😅 This is awesome and needs more visiblity. I completely missed Frida's JDWP implementation when looking at this kind of stuff. Is there anything that stops frida from doing JDWP injection if I invoke `frida -FU`? I guess it requires a bit of plumbing (checking if the target PID is in the JDWP PID list), but having non-root debugging to just work would be really fantastic.

mhils:
Thanks @oleavr! I'm trying to solve the very same "very cool feature but not properly exposed/documented" problem for some mitmproxy parts at the moment, as much as I'd like to play around with Vala I'm afraid I won't find any cycles for this anytime soon. Thank you for the offer! 😊

PR #98, opened 2022-03-12
Use shlex instead of str.split for magic commands

yotamN:
The shlex module is more suitable for shells and let us support quoting
and escaping without any extra work. It's important to have that now to
support paths with spaces in them.


oleavr:
Thanks!

PR #111, opened 2022-08-01
Fix await with no pause

pachoo:
Updated start logic so an application spawned via '-W'/'--await' will honor the '--no-pause' flag.


oleavr:
Thanks!

PR #118, opened 2022-09-24
Add support for multiple input lines

yotamN:
[Kooha-2022-09-24-20-17-49.webm](https://user-images.githubusercontent.com/5249239/192112281-dad07afa-5949-44f6-aa0f-868971c4ce46.webm)


yotamN:
I just need to test the build process on Windows and it will be done

.github/workflows/package.yml
```diff
@@ -0,0 +1,97 @@
+name: package
+on: [push, pull_request]
+jobs:
+  linux_wheels:
```
oleavr, line 4:
```suggestion
  linux-wheels:
```

(Here and elsewhere, for consistency.)

.github/workflows/package.yml
```diff
@@ -0,0 +1,97 @@
...
+    strategy:
+      matrix:
+        arch: [x86, x86_64, arm64, armhf, mips]
+    runs-on: ubuntu-latest
+    container: ghcr.io/frida/x-tools-linux-${{ matrix.arch }}:latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - run: cd agents/tracer && npm install && npm run build
```
oleavr, line 14: Also need to build the other agent. Let's just run `make` here instead, which takes care of both.

.github/workflows/package.yml
```diff
@@ -0,0 +1,97 @@
...
+            exit 1
+          else
+            mv $bdist_wheel_py.patched $bdist_wheel_py
+          fi
+          export _PYTHON_HOST_PLATFORM=linux-$py_arch
+          echo "plat_name = $py_platform" >> setup.cfg
+          pip wheel -w ./dist --no-deps .
+      - uses: actions/upload-artifact@v3
+        with:
+          name: frida_tools_linux
```
oleavr, line 58:
```suggestion
          name: frida-tools-linux
```

Might be better to use a separate artifact per arch though so people can download just the one they need? At least that's how I've done it for frida-python's CI.

.github/workflows/package.yml
```diff
@@ -0,0 +1,97 @@
...
+        with:
+          submodules: recursive
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: cd agents/tracer && npm install && npm run build
+      - run: python3 -m pip install wheel
+      - run: python3 -m pip install tree_sitter
+      - run: python3 setup.py bdist_wheel --plat-name=macosx_10_9_x86_64
+      - run: python3 setup.py bdist_wheel --plat-name=macosx_11_0_x86_64
```
oleavr, line 74:
```suggestion
      - run: python3 setup.py bdist_wheel --plat-name=macosx_11_0_arm64
```

Right? Or perhaps not specifying the platform name is better, as it might build a universal binary.

.github/workflows/package.yml
```diff
@@ -0,0 +1,97 @@
...
+        with:
+          submodules: recursive
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: cd agents/tracer && npm install && npm run build
+      - run: python3 -m pip install wheel
+      - run: python3 -m pip install tree_sitter
+      - run: python3 setup.py bdist_wheel --plat-name=win_amd64
+      - run: python3 setup.py bdist_wheel --plat-name=win32
```
oleavr, line 93: Does this actually work or does it package a 64-bit .DLL in both cases?

.github/workflows/package.yml
yotamN: Actually I don't build arm64, I will remove it

.github/workflows/package.yml
yotamN: Yeah, it's probably 64 bit so I will remove it, don't think it's a common platform

.github/workflows/package.yml
yotamN: Fixed

.github/workflows/package.yml
yotamN: There is no Makefile in there, so I just added the other agent

.github/workflows/package.yml
yotamN: Fixed

.github/workflows/package.yml
oleavr: There's one in the root of the repo that takes care of both agents.

.github/workflows/package.yml
oleavr: I think Python's build system will build a universal binary automatically. There's no need for an Apple Silicon system to cross-compile for arm64. So I think here by just removing `--plat-name` and doing one build, it will build it for both.

.github/workflows/package.yml
oleavr: Right! We shouldn't specify the platform name here, the default should be fine. But what we can do is do setup-python twice, specifying the architecture each time. So that could be a matrix with the two architectures, and that's all we need to support both 32- and 64-bit.

.github/workflows/package.yml
yotamN: Done

.github/workflows/package.yml
yotamN: Done

.github/workflows/package.yml
yotamN: Done

setup.py
```diff
@@ -9,6 +9,8 @@
 
 agents = glob.glob(os.path.join(pkg_dir, "*_agent.*"))
 assert len(agents) > 0, "Agents not compiled; run “npm install && npm run build” in agents/*/"
+os.system("python ./build_aux/build_treesitter.py")
+package_data = agents + ["treesitter.so"]
```
oleavr, line 13: I think we should tie into the build system so we don't do this when just running setup.py to query for something.

.github/workflows/package.yml
```diff
@@ -0,0 +1,95 @@
...
+      - run: make
+      - run: python3 -m pip install wheel
+      - run: python3 -m pip install tree_sitter
+      - run: python3 setup.py bdist_wheel
+      - uses: actions/upload-artifact@v3
+        with:
+          name: frida-tools-macos
+          path: dist/frida_tools-*.whl
+
+  windows-wheels:
```
oleavr, line 79: Would be good to add a matrix here to build for both 32- and 64-bit. The `setup-python` action supports specifying the architecture, so that's everything needed.

.github/workflows/package.yml
```diff
@@ -0,0 +1,95 @@
+name: package
+on: [push, pull_request]
+jobs:
```
oleavr, line 3:
```suggestion

jobs:
```

(Maybe, to make things more "balanced.")

.github/workflows/package.yml
```diff
@@ -0,0 +1,95 @@
...
+  linux-wheels:
```
oleavr, line 4: For consistency with other CI it would be good to order the jobs: Windows, macOS, Linux.

.github/workflows/package.yml
```diff
@@ -0,0 +1,95 @@
...
+    strategy:
+      matrix:
+        arch: [x86, x86_64, arm64, armhf, mips]
```
oleavr, line 7: For consistency with other CI it would be good to order the archs like in the frida repo.

oleavr (changes requested):
Nice! I guess the only missing piece of the puzzle now is CI for publishing to PyPI. Other than that, just some nitpicks:

setup.py
yotamN: I think this is the only way that a source distribution would build correctly?

yotamN:
The cross compilation fix was a bit dirty but it seems this is what [crossenv](https://crossenv.readthedocs.io/en/latest/intro.html#how-it-works) is doing as well.

PR #125, opened 2022-10-27
frida-tools package is not zip_safe

ChiChou:
tracer.py uses relative path to locate `tracer_agent.js`. If the package has been installed as zip (.egg), `frida-trace`  can't read from the tracer template, thus result in failure.

To reproduce, clone this repo and run `python3 setup.py install` on a clean machine (that has not installed `frida-tools` before), then use `frida-trace` command. Then it will complain `[Errno 20] Not a directory: ...tracer_agent.js`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
         "pygments >= 2.0.2, < 3.0.0",
     ],
     license="wxWindows Library Licence, Version 3.1",
-    zip_safe=True,
+    zip_safe=False,
     keywords="frida debugger dynamic instrumentation inject javascript windows macos linux ios iphone ipad android qnx",
     classifiers=[
         "Development Status :: 5 - Production/Stable",


yotamN:
Before merging it, I wonder if we can maybe it zip safe instead

ChiChou:
@yotamN then you have to add extra checks in to read from zip (.egg) format instead of constructing paths from `__file__`

https://github.com/frida/frida-tools/blob/f0867e206f7b9127940c18e1a985626223c03c69/frida_tools/ls.py#L40
https://github.com/frida/frida-tools/blob/f0867e206f7b9127940c18e1a985626223c03c69/frida_tools/push.py#L50
https://github.com/frida/frida-tools/blob/f0867e206f7b9127940c18e1a985626223c03c69/frida_tools/pull.py#L62
https://github.com/frida/frida-tools/blob/f0867e206f7b9127940c18e1a985626223c03c69/frida_tools/tracer.py#L343
https://github.com/frida/frida-tools/blob/f0867e206f7b9127940c18e1a985626223c03c69/frida_tools/rm.py#L43

yotamN:
It seems that [the property is deprecated](https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html) so I don't think it's worth refactoring everything to be zip safe.  
Keeping the property is fine I guess, so I will merge.

PR #134, opened 2023-02-19
Update The frida-create Agent Template

oriori1703:
Update the npm dependencies in the agent template, and change the TS module resolution setting to reflect Frida's use of ESM


oleavr:
Thanks! 🙌 (Apologies for the delay here, that was not intentional.)

PR #135, opened 2023-02-27
binaryxml: fix stringpool insertion padding

mhils:
I previously had some APKs that failed after patching, this fixes it. 😃

### DIFF
--- a/frida_tools/apk.py
+++ b/frida_tools/apk.py
@@ -264,7 +264,7 @@ class StringPool:
 
         string = None
         if self.utf8:
-            # Ignore UTF-16 length
+            # Ignore number of characters
             n = struct.unpack("<B", self.stream.read(1))[0]
             if n & 0x80:
                 n = ((n & 0x7F) << 8) | struct.unpack("<B", self.stream.read(1))[0]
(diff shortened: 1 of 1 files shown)


frida_tools/apk.py
```diff
@@ -299,10 +299,11 @@ def append_str(self, add: str) -> int:
...
 
         if self.utf8:
-            # UTF-16 length (ignored)
-            chunk_data.extend(struct.pack("<B", len(add)))
-            # UTF-8 length
+            assert len(add.encode("utf-8")) < 128  # multi-byte len strings not supported yet
+            # the length of the string in characters
             chunk_data.extend(struct.pack("<B", len(add)))
+            # the length of the UTF-8 encoding of the string in bytes
+            chunk_data.extend(struct.pack("<B", len(add.encode("utf-8"))))
```
oleavr, line 306:
```suggestion
            length_in_characters = len(add)
            length_in_bytes = len(add.encode("utf-8"))
            chunk_data.extend(struct.pack("<BB", length_in_characters, length_in_bytes))
```

(Should avoid comments if the code can be expressed such that they're not needed.)

frida_tools/apk.py
```diff
@@ -313,6 +314,9 @@ def append_str(self, add: str) -> int:
             # Insert a UTF-16 NUL
             chunk_data.extend([0, 0])
 
+        # pad to a multiple of 4 bytes
+        chunk_data.extend([0] * (4 - len(chunk_data) % 4))
```
oleavr, line 318:
```suggestion
        alignment_padding = [0] * (4 - len(chunk_data) % 4)
        chunk_data.extend(alignment_padding)
```

oleavr: Won't this add 4 bytes of padding when `chunk_data` is already a multiple of four?

oleavr (changes requested):
Ahh nice, thanks! 🙌 (And apologies for the absurd delay, finally now starting to catch up on things after life got super-busy.)

frida_tools/apk.py
mhils: It did, fixed! 😃

oleavr:
Thanks! 🙌

PR #136, opened 2023-03-07
Fix folders/files name generation inside __handlers__

diaul:
If you use frida-trace. on native functions inside specific Android application shared objects, the files and folders inside __handlers__,  are created with the wrong names. This patch fix this issue.


oleavr:
Thanks! 🙌

PR #137, opened 2023-03-18
Fix Stack Trace Trim

oriori1703:
Fix an duplicate lines in multiline error messages.

for example:

```js
throw Error("test1\ntest2\ntest3")
```
The code above would result in:
```
Error: test1
test2
test3
test2
test3
    at <eval> (<input>:1)
    at eval (native)
```

I also removed the last line in the stack trace, so the now the same code would result in:
```
Error: test1
test2
test3
    at <eval> (<input>:1)
```


yotamN:
LGTM

PR #138, opened 2023-04-20
repl: Detach session before killing on exit

mrmacete:
This is to avoid random hang on exit observed when using `--kill-on-exit` when spawning iOS apps, which after hitting CTRL+C results in:
```
Frida:ERROR:../../../frida-core/src/frida.vala:1409:_frida_device_release_session_co: assertion failed: (session_id != null)
Bail out! Frida:ERROR:../../../frida-core/src/frida.vala:1409:_frida_device_release_session_co: assertion failed: (session_id != null)
Abort trap: 6
```


oleavr:
Thanks! 🤘

Issue #148, opened 2024-01-23
respond to the NO_COLOR environment variable by suppressing color output


smcpeak:
The Frida console tools such as `frida-trace` should, IMO, respond to the `NO_COLOR` environment variable.  Specifically, when this variable is set to a non-empty string, the tools should not emit color-changing escape sequences.  See https://no-color.org/ .

I believe the following diff suffices:

```
--- a/frida_tools/application.py
+++ b/frida_tools/application.py
@@ -147,7 +147,10 @@ class ConsoleApplication:
         if hasattr(signal, "SIGPIPE"):
             signal.signal(signal.SIGPIPE, signal.SIG_DFL)
 
-        colorama.init(strip=True if plain_terminal else None)
+        # If true, emit text without colors.  https://no-color.org/
+        no_color = plain_terminal or bool(os.environ.get("NO_COLOR"))
+
+        colorama.init(strip=True if no_color else None)
 
         parser = self._initialize_arguments_parser()
         real_args = compute_real_args(parser, args=args)
```

Note that the changed code does *not* set `plain_terminal` (and hence `self._plain_terminal`) to `True` if `NO_COLOR` is set, because `self._plain_terminal` is used by `ps.py` and `repl.py` for terminal capabilities unrelated to color, and `NO_COLOR` should not affect those.

I'd be happy to submit a PR if that would be helpful.

s1341:
Please submit a PR.

PR #149, opened 2024-01-24
application.py: Respond to NO_COLOR environment variable.

smcpeak:
When `NO_COLOR` is set to a non-empty string, this disables emission of colored text, per the specification proposal at https://no-color.org/ .

In the modified code, setting `NO_COLOR` does *not* set the `plain_terminal` flag because that flows into `self._plain_terminal`, which is used elsewhere for things unrelated to colors.

Fixes frida/frida-tools#148.


oleavr:
Thanks! (And apologies for the delay!)

PR #156, opened 2024-05-16
skip typescript lint for generated code

ChiChou:
when using IDE for this template, the linter also checks generated javascript, giving many non-sense errors like duplicated variable names

exclude generated code to tell the linter to ignore it


oleavr:
Thanks!

Issue #158, opened 2024-05-21
frida-tools 12.4.2 wheel build fails


seanthegeek:
OS: REMnux 2024.21.5 (Based on Ubuntu 20.04)
Python: 3.8

```
sudo pip3 install -U frida-tools
Requirement already satisfied: frida-tools in /usr/local/lib/python3.8/dist-packages (12.3.0)
Collecting frida-tools
  Using cached frida-tools-12.4.2.tar.gz (345 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "/usr/local/lib/python3.8/dist-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/usr/local/lib/python3.8/dist-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/usr/local/lib/python3.8/dist-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
        File "/tmp/pip-build-env-jtgnfn44/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 325, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
        File "/tmp/pip-build-env-jtgnfn44/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 295, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-jtgnfn44/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 487, in run_setup
          super().run_setup(setup_script=setup_script)
        File "/tmp/pip-build-env-jtgnfn44/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 311, in run_setup
          exec(code, locals())
        File "<string>", line 99, in <module>
      TypeError: 'type' object is not subscriptable
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

[...]

PR #159, opened 2024-05-29
setup: Fix python<3.9 list type

UndefinedIdentifier:
fix this [issue](https://github.com/frida/frida-tools/issues/158)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,7 @@ import os
 import shutil
 import sys
 from pathlib import Path
-from typing import Iterator
-from typing import List
+from typing import Iterator, List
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -3,6 +3,7 @@
 import sys
 from pathlib import Path
 from typing import Iterator
+from typing import List
```
oleavr, line 6:
```suggestion
from typing import Iterator, List
```

oleavr (changes requested):
Thanks, great catch!

oleavr:
(Just released 12.4.3 with this fix 🎉)

PR #163, opened 2024-09-22
Fixed missing defineHandler in _create_stub_java_handler.

### LINKED ISSUE #162: Frida-trace Parser Error On Loaded Java Method Handlers JS Files.
On the latest version of frida-tools, frida-trace generates `Warning: /handlers/$init.js: could not parse '/handlers/$function.js' line 16: expecting ';'` when used to trace Java methods with the -j flag. It does not throw the error with other flags. Going through the frida_tools\tracer.py and juxtaposing the generated handlers js files for a Java method trace using -j and a "str*" trace using -i, the `def _create_stub_java_handler(self, target: TraceTarget, decorate)` in the tracer.py returns a block of code with incorrect syntax for both QJS and V8 engs. It is apparently missing a function wrapper before the first "{".

Manually appending a "defineHandler();" wrapper for each generated Java handler file fixes this, as in:

> 
> /*
> 
>     Auto-generated by Frida. Please modify to match the signature of %(display_name)s.
>     For full API reference, see: https://frida.re/docs/javascript-api/
>     */
> 
> defineHandler({
> /**
> 
>     Called synchronously when about to call %(display_name)s.
>     @this {object} - The Java class or instance.
>     @param {function} log - Call this function with a string to be presented to the user.
>     @param {array} args - Java method arguments.
>     @param {object} state - Object allowing you to keep state across function calls.
>     */
>     onEnter(log, args, state) {
>     log(%(display_name)s(${args.map(JSON.stringify).join(', ')}));
>     },
> 
> /**
> 
>     Called synchronously when about to return from %(display_name)s.
>     See onEnter for details.
>     @this {object} - The Java class or instance.
>     @param {function} log - Call this function with a string to be presented to the user.
>     @param {NativePointer} retval - Return value.
>     @param {object} state - Object allowing you to keep state across function calls.
>     */
>     onLeave(log, retval, state) {
>     if (retval !== undefined) {
>     log(<= ${JSON.stringify(retval)});
>     }
>     }
>     });

Kindly make this adjustment at the code-generation level, thanks and great work!

kokoslik:
Fixes #162 by adding missing "defineHandler" into the _create_stub_java_handler.

### DIFF
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -792,7 +792,7 @@ defineHandler({
  * For full API reference, see: https://frida.re/docs/javascript-api/
  */
 
-{
+defineHandler({
   /**
    * Called synchronously when about to call %(display_name)s.
    *
@@ -820,7 +820,7 @@ defineHandler({
       log(`<= ${JSON.stringify(retval)}`);
     }
   }
-}
+});
 """ % {
             "display_name": target.display_name
         }


oleavr:
Thanks, great catch! 🙌

PR #170, opened 2024-10-18
tracer: Bind UI to a random port

ChiChou:
Let the OS to choose an idle port instead of iteration and retry

### DIFF
--- a/frida_tools/tracer.py
+++ b/frida_tools/tracer.py
@@ -5,7 +5,6 @@ import asyncio
 import binascii
 import codecs
 import email.utils
-import errno
 import gzip
 import http
 import mimetypes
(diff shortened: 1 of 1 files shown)


ChiChou:
But there seems to be a race condition...

`self._ui_port` is shared between two threads, we can not guarantee `on_trace_progress` happends after websocket server initialization

oleavr (changes requested):
Thanks! Could you please reformat the commit messages to follow the existing style? E.g. `tracer: Capitalized summary`, loosely based on https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html.

oleavr:
> But there seems to be a race condition...
> 
> `self._ui_port` is shared between two threads, we can not guarantee `on_trace_progress` happends after websocket server initialization

Glad you saw this. I did this intentionally as a temporary hack, but then forgot to fix it :sweat_smile:

PR #182, opened 2025-04-15
ps: also apply -e(exclude icon) option on list-application

RuffaloLavoisier:


RuffaloLavoisier:
@oleavr 
My mistake, apply -e option on list-application.

RuffaloLavoisier:
@oleavr 
Could you please check it?

PR #183, opened 2025-05-31
frida-create: upgrade project template to frida 17

ChiChou:
The template was not updated for breaking changes from frida@17


ChiChou:
Note: do not update `@types/node` yet, there will be a global name conflict `File`

ChiChou:
Related issue in frida-compile: https://github.com/frida/frida-compile/pull/100

PR #184, opened 2025-06-01
Prevent treeshake of frida-objc-bridge

NSEcho:
This PR fixes an issue where `objc_msgSend` is dropped in the process of treeshaking where it sees `objc_msgSend` unused. The solution to this is disabling treeshake of `frida-objc-bridge` altogether.

This issue correctly fixes https://github.com/frida/frida-objc-bridge/issues/64 along with https://github.com/frida/frida/issues/3460.

Additionally, fixes https://github.com/frida/frida/issues/3480 which are all basically the same.

### DIFF
--- a/bridges/rollup.config.ts
+++ b/bridges/rollup.config.ts
@@ -27,7 +27,7 @@ export default defineConfig(BRIDGES.map(name => {
                         return {
                             code,
                             map: null,
-                            moduleSideEffects: 'no-treeshake',
+                            moduleSideEffects: "no-treeshake",
                         };
                     }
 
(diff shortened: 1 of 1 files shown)


Auties00:
> Thanks! 🙌

Would it be possible to release this change? 
I was reversing a feature on iOS, randomy updated frida and now I'm having this problem.
I can always compile the frida-server and spin it up manually on the phone, but as I'm going to sleep, I figured I might as well ask, so tomorrow I don't have to do it,
I bet others are in this situation lol.
In any casse thanks for all your work, and to all the mantainers as well.

PR #185, opened 2025-06-02
Fix the `frida-create` template

oriori1703:
Fix multiple issues in the template used by `frida-create`

1. It seems typescript doesn't like the changes to tsconsfig's `moduleResolution` in https://github.com/frida/frida-tools/pull/183
I throw the following error:
```
tsconfig.json:8:25 - error TS5109: Option 'moduleResolution' must be set to 'Node16' (or left unspecified) when option 'module' is set to 'Node16'.
```

2. Imports are broken because the package is not specified as an ESModule in the `package.json`.
This causes the following error:
```
ReferenceError: 'exports' is not defined
    at <anonymous> (/agent/index.js:1)
```

### DIFF
--- a/frida_tools/creator.py
+++ b/frida_tools/creator.py
@@ -73,6 +73,7 @@ class CreatorApplication(ConsoleApplication):
   "version": "1.0.0",
   "description": "Frida agent written in TypeScript",
   "private": true,
+  "type": "module",
   "main": "agent/index.ts",
   "scripts": {{
     "prepare": "npm run build",
@@ -92,14 +93,14 @@ class CreatorApplication(ConsoleApplication):
         ] = """\
 {
   "compilerOptions": {
-    "target": "es2020",
-    "lib": ["es2020"],
+    "target": "es2022",
+    "lib": ["es2022"],
     "allowJs": true,
     "noEmit": true,
     "strict": true,
     "esModuleInterop": true,
-    "moduleResolution": "node",
-    "module": "esnext"
+    "moduleResolution": "Node16",
+    "module": "Node16"
   },
   "exclude": ["_agent.js"]
 }


ChiChou:
But yes I forgot to put type: module there

oleavr:
Thanks! :raised_hands:

oleavr:
Just released frida-tools 14.1.0 with this included. Would recommend trying it together with Frida 17.1.0, also just released -- the new Compiler in 17.1.0 makes compilation blazing fast thanks to ESBuild + typescript-go. There's also an option to disable type-checking, which means only ESBuild is used. Plus, we now support customizing the output and bundle formats used.

oriori1703:
> Just released frida-tools 14.1.0 with this included. Would recommend trying it together with Frida 17.1.0, also just released -- the new Compiler in 17.1.0 makes compilation blazing fast thanks to ESBuild + typescript-go. There's also an option to disable type-checking, which means only ESBuild is used. Plus, we now support customizing the output and bundle formats used.

That's sounds really cool!
I didn't know typescript-go was already stable enough 😯

oleavr:
> > Just released frida-tools 14.1.0 with this included. Would recommend trying it together with Frida 17.1.0, also just released -- the new Compiler in 17.1.0 makes compilation blazing fast thanks to ESBuild + typescript-go. There's also an option to disable type-checking, which means only ESBuild is used. Plus, we now support customizing the output and bundle formats used.
> 
> That's sounds really cool! I didn't know typescript-go was already stable enough 😯

Glad to hear :relaxed: It's not yet on parity with the legacy codebase, but the areas it's lacking in aren't needed for our use-case.

Btw, just tweaked frida-create some more, so the Node.js-based frida-compile isn't part of the template, and made one more tweak to tsconfig.json -- both changes now available in frida-tools 14.1.1.

PR #186, opened 2025-06-06
Ci/wheels

oriori1703:
Build and publish frida-tools in the CI

This is a continuation of https://github.com/frida/frida/pull/3482

fixes: https://github.com/frida/frida-tools/issues/176


oriori1703:
@oleavr This supposedly works, but I haven't checked that the upload actually works, because you have to setup [trusted publishing](https://docs.pypi.org/trusted-publishers/adding-a-publisher/).

Also you might want to change when those jobs run

oriori1703:
Also I noticed that `frida-python` is specified as dependency in meson, which means that we have to build `frida-python` (and all of its deps) just to build this one.

IMO because `frida-python` is a runtime dependency it doesn't really fit in meson.
Maybe we should remove it, and just rely on pip for managing that dependency?

oriori1703:
Hey @oleavr, just a gentle ping on this PR when you have time. Would appreciate your review! :smile: 

Also I saw you switched back to setuptools in e7a4b28. 
I didn't really understand what was the issue your were having, but if you want I would be happy to take a look with a bit more info.

PR #187, opened 2025-06-06
Migrate from setuptools to meson-python

oriori1703:
While working on [publishing wheels](https://github.com/frida/frida-tools/pull/186) I realised the build could be modernised and simplified by using meson-python.

This allows us to use more modern packaging standards (like pyproject.toml) and fits well with the existing meson build system.

TODO:
- [x] ~Fix building with frida-python (unless we remove it as suggested in a [previous comment](https://github.com/frida/frida-tools/pull/186#issuecomment-2949517672))~ Removed the dependency for now 
- [x] Fix version detection. For now I just hardcoded it in the `meson.build` 
- [x] Fix license not being recognised by the build system


oleavr:
Nice! The only issue I can see with this approach is that we have a Meson wrapper that takes care of downloading prebuilt tools + libraries. Not sure we want to lose this.

oriori1703:
> Nice! The only issue I can see with this approach is that we have a Meson wrapper that takes care of downloading prebuilt tools + libraries. Not sure we want to lose this.

As far as I understand, all of the libraries and tools that are downloaded are not really need when developing frida-tools.
Also, if you still think it is useful to have those tools you can just direct people to the [frida/frida](https://github.com/frida/frida-tools/pull/187) repo which functions the same way (download all of the related projects, and all the tools needed to build them).

I have fixed all other issues, so if you agree with removing frida-python as a dependency, everything should be ready to merge :smile:

